PR #2, opened 2025-07-21
Support trade listener 

aaxxios:
Add support to listen to trade events when adding orders


joaquinbejar:
Looks good now—tests pass on my side too!
Thanks for tightening up the docs and handling the invalid-ID case. I’m happy with the changes, so I’ll go ahead and merge. 🚀

PR #3, opened 2025-07-25
fix canceller_thread use random order_id to cancel non-existent order in HFT simulation

SYSUljz:
The previous HFT simulation had a  flaw in the cancellation testing - it was generating random UUIDs that would never match actual orders in the order book, making cancellation testing ineffective and unrealistic.


joaquinbejar:
@SYSUljz Thanks for fixing the cancellation path—this makes the HFT sim more realistic and resolves the previous random-UUID issue.

PR #34, opened 2025-10-28
Price Level Notifications

aaxxios:
This PR adds full support for price level change events across Adding, Cancelling, Updating and Matching orders.

Users can now hook into the notifications by setting price level listener on the order book. This provides an entry point to  react to book changes and update external applications such as UI order books


joaquinbejar:
Thx for the contribution, LGTM

Issue #97, opened 2026-06-23
fix(matching): partial-fill STP CancelTaker/CancelBoth rests the residual instead of cancelling it


joaquinbejar:
## Overview

match_order_inner only records the taker's terminal Cancelled{SelfTradePrevention} and returns the STP error when stp_taker_cancelled AND no_fills. When non-self fills already occurred (safe_quantity>0, or fills at earlier levels), it returns Ok with remaining_quantity>0 and no STP-cancel signal on MatchResult. The caller then rests the GTC remainder in the book (or mislabels an IOC as InsufficientLiquidity), so CancelTaker/CancelBoth fail to remove the residual taker after a partial self-cross.

**Why this matters.** Reproduced: a GTC taker that partially fills against another user and then hits its own resting order under CancelTaker leaves its residual resting as a live order instead of being cancelled — directly defeating the STP mode and never recording the SelfTradePrevention terminal reason. This is a correctness/safety bug in self-trade prevention, gated behind a partial-fill self-cross path.

> Surfaced by an automated multi-agent code review (adversarially verified). Severity: **high** · Effort: **M** · Category: **correctness**.

> ⚠️ **Gated — needs a design decision before implementing:** design decision

## Context

Affected files:

- `src/orderbook/matching.rs`
- `src/orderbook/modifications.rs`

See `rules/global_rules.md` and `CLAUDE.md` for the binding conventions.

## Proposed approach

Carry an explicit stp_taker_cancelled flag on MatchResult (or return a dedicated Ok variant) so the caller in add_order does not rest the remainder when STP cancelled the taker, regardless of whether fills occurred. Record OrderStatus::Cancelled{ filled_quantity, reason: SelfTradePrevention } for the taker with the actual filled quantity on the partial-fill path. Add a unit test: GTC buy partially fills against another user, then hits its own resting ask under CancelTaker — assert the residual is NOT rested and the taker terminal state is Cancelled{SelfTradePrevention}.

## Acceptance criteria

- [ ] A partially-filled taker that STP cancelled does not rest its remainder in the book (GTC) under CancelTaker/CancelBoth
- [ ] The taker records OrderStatus::Cancelled{ reason: SelfTradePrevention } with the true filled quantity on partial-fill self-cross
- [ ] MatchResult exposes the STP-cancel decision to the caller
- [ ] New test asserts no residual is rested for a partial-fill self-cross
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean
[...]

PR #184, opened 2026-06-26
feat(modifications): add method that returns trade result immediately after adding order to book

Dev380:
Currently the orderbook only supports `add_order` or `match_order` to add orders to the book, excluding convenience functions. This makes the API slightly annoying to use in certain exchange designs because to get the result of an order, you either have to use `match_order` and give up the flexibility of using any `OrderType` object, or you have to use `add_order` and wait for the listener to process it, which may not be ideal especially if you want to read the trade result in the same function that the order is added. This adds an `add_order_with_result` method that also returns the TradeResult and makes `add_order` a simple wrapper around it for compatibility.

### DIFF
--- a/src/orderbook/modifications.rs
+++ b/src/orderbook/modifications.rs
@@ -1017,11 +1017,30 @@ where
 
     /// Add a new order to the book, automatically matching it if it's aggressive.
     ///
+    /// This convenience method calls [`Self::add_order_with_result`] internally but
+    /// ignores the trade result in the output.
+    ///
     /// # Errors
     /// Returns [`OrderBookError::KillSwitchActive`] when the kill switch
     /// is engaged. The check runs before any cache invalidation, STP
     /// validation, tick/lot validation, or matching work.
-    pub fn add_order(&self, mut order: OrderType<T>) -> Result<Arc<OrderType<T>>, OrderBookError> {
+    pub fn add_order(&self, order: OrderType<T>) -> Result<Arc<OrderType<T>>, OrderBookError> {
+        self.add_order_with_result(order).map(|(order, _)| order)
+    }
+
+    /// Add a new order to the book, automatically matching it if it's aggressive, also
+    /// returning the trade result directly.
+    ///
+    /// This method also calls the trade listener if enabled in addition to returning the trade result.
+    ///
+    /// # Errors
+    /// Returns [`OrderBookError::KillSwitchActive`] when the kill switch
+    /// is engaged. The check runs before any cache invalidation, STP
+    /// validation, tick/lot validation, or matching work.    
+    pub fn add_order_with_result(
+        &self,
+        mut order: OrderType<T>,
+    ) -> Result<(Arc<OrderType<T>>, Option<TradeResult>), OrderBookError> {
         self.check_kill_switch_or_reject(order.id())?;
         // Pre-trade risk gate: per-account open-orders / notional /
         // price band. No-op when no `RiskConfig` is installed.
@@ -1096,20 +1115,6 @@ where
             order.user_id(),
         )?;
 
-        let trades_emitted = match_result.trades().len() as u64;
-        if trades_emitted > 0 {
-            crate::orderbook::metrics::record_trades(trades_emitted);
-            if let Some(ref listener) = self.trade_listener {
-                let mut trade_result = TradeResult::with_fees(
-                    self.symbol.clone(),
-                    match_result.clone(),
-                    self.fee_schedule,
-                );
-                trade_result.engine_seq = self.next_engine_seq();
-                listener(&trade_result) // emit trade events to listener
-            }
-        }
-
         // True (non-self) executed quantity. `remaining_quantity` only decrements on
         // real trades, so STP-prevented self-fills never count toward it.
         let original_qty = order.total_quantity();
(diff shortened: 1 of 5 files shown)


src/orderbook/modifications.rs
```diff
@@ -1135,6 +1140,24 @@ where
             });
         }
 
+        let trades_emitted = match_result.trades().len() as u64;
+        let trade_result = if trades_emitted > 0 {
```
joaquinbejar: 🔴 **Critical**: Moving this block below the `taker_stp_cancelled` early return changes behavior on the STP partial-fill path from #97. When a taker gets real fills at earlier levels and STP then cancels the residual, those fills previously reached the trade listener and `record_trades`. Now the function returns `SelfTradePrevented` before this block runs, so trades that actually executed against resting makers never hit the event stream or the metrics. Any downstream consumer of the trade feed will miss real executions. The emission has to happen before that early return, or the STP branch needs its own emission before returning. CI is green because no test covers the listener on that path, which is exactly why this needs one.

src/orderbook/modifications.rs
```diff
@@ -1135,6 +1140,24 @@ where
...
+            crate::orderbook::metrics::record_trades(trades_emitted);
+            let mut trade_result = TradeResult::with_fees(
+                self.symbol.clone(),
+                match_result.clone(),
```
joaquinbejar: 🟠 **Warning**: This now builds a `TradeResult` unconditionally for every matched order, so plain `add_order` callers with no listener installed pay a `symbol.clone()` plus a full `match_result.clone()` per aggressive order that they previously didn't. That is an allocation regression on the matching hot path. Worth gating construction on `self.trade_listener.is_some()` or an internal flag for whether the caller wants the result, so the old `add_order` cost is preserved.

src/orderbook/modifications.rs
```diff
@@ -1017,11 +1017,30 @@ where
...
+    /// Add a new order to the book, automatically matching it if it's aggressive, also
+    /// returning the trade result directly.
+    ///
+    /// This method also calls the trade listener if enabled in addition to returning the trade result.
+    ///
+    /// # Errors
+    /// Returns [`OrderBookError::KillSwitchActive`] when the kill switch
+    /// is engaged. The check runs before any cache invalidation, STP
+    /// validation, tick/lot validation, or matching work.    
+    pub fn add_order_with_result(
```
joaquinbejar, line 1058: 🟠 **Warning**: New public method lands with no tests and no CHANGELOG entry. It needs at least a unit test asserting the returned `TradeResult` matches what the listener saw for a partial fill and a full fill, plus a case asserting `None` when nothing matched, and one covering the STP partial-fill interaction mentioned above.

src/orderbook/modifications.rs
```diff
@@ -1017,11 +1017,30 @@ where
...
+        &self,
+        mut order: OrderType<T>,
+    ) -> Result<(Arc<OrderType<T>>, Option<TradeResult>), OrderBookError> {
```
joaquinbejar, line 1073: 💡 **Suggestion**: On the IOC partial-fill path this still returns `Err(InsufficientLiquidity)` and the freshly built trade result is dropped, so the headline use case of reading fills where the order was placed still does not work for IOC takers. Same on the STP partial-fill error path. Worth either carrying the result inside those error variants or documenting clearly that fills are only observable via the listener when the residual gets cancelled.

joaquinbejar (dismissed):
Thanks for this, the API gap you describe is real and returning the trade result from the add path is a good idea. But moving the emission block below the STP early return silently drops real fills from the trade stream on the partial-fill STP path, and the unconditional TradeResult construction adds clones to the hot path for callers that never asked for the result. It also needs tests and a changelog entry for the new public method. REQUEST_CHANGES.

joaquinbejar:
Since this has been quiet for a few days I pushed the requested changes myself (549df7d) so we can land it — thanks again for the contribution, the API gap was real.

What changed on top of your commit:

- **Trade emission moved back above the STP taker-cancel early return.** In e27874b the emission block ran after the `taker_stp_cancelled` early return, so a taker that partially filled against another user and was then STP-cancelled dropped those real fills from `record_trades` and the trade listener. Regression tests added for both `add_order` and `add_order_with_result`.
- **`TradeResult` construction is now gated.** Both public methods delegate to a private `add_order_inner(order, want_result)`; the result is built only when `want_result || trade_listener.is_some()`, so the plain `add_order` path with no listener keeps main's allocation profile (no extra `MatchResult` clone, no extra `engine_seq` tick).
- **Docs**: clarified `None` semantics, error-path behavior (IOC remainder / STP cancel after fills reach the listener only), and `engine_seq` consumption of the new method.
- **Tests**: 7 new cases (full fill, resting → `None`, partial fill remainder, listener/caller parity, STP partial-fill emission ×2, IOC partial-fill emission). CHANGELOG entry and `lib.rs` What's New added, README regenerated.

Will merge once CI is green.

joaquinbejar:
Released in [v0.9.1](https://github.com/joaquinbejar/OrderBook-rs/releases/tag/v0.9.1), published on [crates.io](https://crates.io/crates/orderbook-rs/0.9.1). Thanks for the contribution @Dev380!

PR #198, opened 2026-07-12
Add fallible fee calculation and exact-input bound to FeeSchedule (#197)

### LINKED ISSUE #197: FeeSchedule::calculate_fee saturates — add a fallible variant or a documented exact-input bound
As of `575de34260b0`, `FeeSchedule::calculate_fee` (src/orderbook/fees.rs:111-128) deliberately saturates:

```rust
let magnitude_u128 = notional.saturating_mul(u128::from(bps.unsigned_abs())) / 10_000;
let magnitude = i128::try_from(magnitude_u128).unwrap_or(i128::MAX);
```

The in-code comment documents why this is safe "for realistic inputs" — but a consumer that must *guarantee* exact fees (journaled, replayable venues) has no way to distinguish a saturated fee from an exact one, and no documented input bound to validate against.

## Suggested work (either or both)

- A fallible variant — `try_calculate_fee(&self, notional, is_maker) -> Result<i128, FeeOverflow>` — so venues can reject the order instead of journaling a clamped fee.
- A documented `const` input bound (max notional × max |bps| that stays exact) that callers can enforce at admission time, making the saturating branches provably unreachable.

## Context

Found while designing [fauxchange](https://github.com/joaquinbejar/fauxchange), whose fee replay contract requires exact integer fees. Its current plan is a venue-owned price/quantity/bps bound checked at startup; a fallible upstream API or published bound would make that contract directly enforceable. Related: Option-Chain-OrderBook's `ValidationConfig` has no max-price knob to enforce such a bound (issue filed there).

joaquinbejar:
## Summary

`FeeSchedule::calculate_fee` deliberately saturates when `notional × |bps|` overflows `u128`, so a consumer that must guarantee exact integer fees (journaled, replayable venues such as fauxchange's fee-replay contract) had no way to distinguish a saturated fee from an exact one, and no documented input bound to validate against. This PR ships both remedies suggested in #197: a fallible `try_calculate_fee` that refuses to clamp, and a published exact-input bound enforceable at admission time.

## Changes

- `FeeSchedule::try_calculate_fee(notional, is_maker) -> Result<i128, FeeOverflow>` — identical rounding to `calculate_fee` (truncation toward zero, sign applied after the unsigned-domain magnitude), but returns an error instead of saturating; an `Ok` value is always the mathematically exact `sign(bps) × ⌊notional × |bps| / 10_000⌋`.
- `FeeOverflow` — `thiserror` error carrying the offending `notional`, the signed `bps` that applied, and the `max_exact_notional` for that rate, so venues can reject the order instead of journaling a clamped fee.
- `FeeSchedule::max_exact_notional_for_bps(bps)` (`const fn`) — the exact-input bound: `u128::MAX / |bps|` (`u128::MAX` for a zero rate). `calculate_fee` is exact iff `notional` is at or below it.
- `FeeSchedule::max_exact_notional()` (`const fn`) — minimum over the maker and taker legs; a single venue-level admission bound that makes the saturating branch provably unreachable.
- `calculate_fee` now delegates to `try_calculate_fee` with a bit-identical saturated fallback (magnitude `u128::MAX / 10_000`, signed per `bps`), and its docs gained a `# Saturation` section stating the exactness condition.

## Technical decisions

- **Exactness analysis**: the multiply is the only lossy step — when `notional × |bps|` fits in `u128`, the post-division magnitude is at most `u128::MAX / 10_000 ≈ 3.4e34 < i128::MAX`, so the `i128` conversion can never fail. The saturation predicate is therefore exactly `notional > u128::MAX / |bps|`, which is what the published bound encodes. The (provably unreachable) `i128::try_from` failure in `try_calculate_fee` still maps to `FeeOverflow` so exactness is a checked guarantee rather than a comment.
- **`FeeOverflow` stays standalone** (not aggregated into `OrderBookError`): the engine itself never produces it — `TradeResult::with_fees` keeps the saturating path by design. It exists for external venues with an exact-fee contract.
- **No behavior change to `calculate_fee`**: the delegation preserves the historical output bit-for-bit for every input (locked by the existing saturation tests plus a new clamp test), so journaled fee streams are unaffected. No wire-format change and no `ORDERBOOK_SNAPSHOT_FORMAT_VERSION` bump.

## Public API impact

- New: `FeeOverflow` (re-exported at the crate root alongside `FeeSchedule`), `FeeSchedule::try_calculate_fee`, `FeeSchedule::max_exact_notional_for_bps`, `FeeSchedule::max_exact_notional`.
- No removals or renames. `src/prelude.rs` untouched (`FeeSchedule` is intentionally not in the prelude).
- `src/lib.rs` What's New updated for 0.10.4; `README.md` regenerated via `make readme`; `CHANGELOG.md` entry added; crate version bumped `0.10.3 → 0.10.4`.

## Testing

- [x] Unit tests added or updated (exactness vs `calculate_fee`, overflow error fields including signed maker bps, tight boundary at `max_exact_notional_for_bps ± 1`, zero-bps never errs, `i32::MIN` bps, min-over-legs bound, documented saturated clamp, `Display` content)
- [x] Integration test added in `tests/fee_tests.rs` (admission-style: at or below `max_exact_notional()` both legs are exact and agree with `calculate_fee`; above it the binding leg errs)
- [x] `make pre-push` run locally and clean (1344 tests passed, 0 failed; clippy `-D warnings` clean; release build clean)

## Checklist

- [x] Module boundaries respected (`fees.rs` remains leaf; no deps on `manager`, `nats*`, `sequencer/`)
- [x] No new dependencies added
- [x] `src/lib.rs` docs + `README.md` (via `make readme`) + `CHANGELOG.md` updated

Closes #197

### DIFF
--- a/src/orderbook/fees.rs
+++ b/src/orderbook/fees.rs
@@ -2,6 +2,43 @@
 
 use serde::{Deserialize, Serialize};
 
+/// Denominator for basis-point fee math: 1 bps = 1 / 10_000 of the notional.
+const BPS_DENOMINATOR: u128 = 10_000;
+
+/// Fee computation would exceed the `u128` domain and cannot be exact.
+///
+/// Returned by [`FeeSchedule::try_calculate_fee`] when
+/// `notional × |bps|` overflows `u128` — the only case in which
+/// [`FeeSchedule::calculate_fee`] saturates instead of producing the exact
+/// fee. Venues that must guarantee exact integer fees (journaled, replayable
+/// systems) can reject the input instead of recording a clamped fee, or
+/// enforce [`FeeSchedule::max_exact_notional`] at admission time so this
+/// error is provably unreachable.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+#[error(
+    "fee overflow: notional {notional} × |{bps}| bps exceeds the u128 domain; max exact notional at this rate is {max_exact_notional}"
+)]
+pub struct FeeOverflow {
+    /// The notional value (price × quantity) that was passed in.
+    pub notional: u128,
+    /// The signed fee rate in basis points that applied (maker or taker).
+    pub bps: i32,
+    /// Largest notional that stays exact at this rate — equal to
+    /// [`FeeSchedule::max_exact_notional_for_bps`]`(bps)`.
+    pub max_exact_notional: u128,
+}
+
+impl FeeOverflow {
+    #[cold]
+    fn new(notional: u128, bps: i32) -> Self {
+        Self {
+            notional,
+            bps,
+            max_exact_notional: FeeSchedule::max_exact_notional_for_bps(bps),
+        }
+    }
+}
+
 /// Configurable fee schedule for maker and taker fees
 ///
 /// Fees are expressed in basis points (bps), where 1 bps = 0.01% = 0.0001.
@@ -87,6 +124,17 @@ impl FeeSchedule {
     /// magnitude `7.5015` / `3.0006`, then signed). External fee reconciliation
     /// should therefore truncate-toward-zero, not round-half-up.
     ///
+    /// # Saturation
+    ///
+    /// The result is **exact** if and only if
+    /// `notional <= Self::max_exact_notional_for_bps(bps)` (equivalently,
+    /// `notional × |bps|` fits in `u128`). Beyond that bound the fee clamps
+    /// to a magnitude of `u128::MAX / 10_000` (signed per `bps`) instead of
+    /// panicking. Callers that must distinguish a saturated fee from an
+    /// exact one should use [`Self::try_calculate_fee`], or enforce
+    /// [`Self::max_exact_notional`] at admission time so the saturating
+    /// branch is provably unreachable.
+    ///
     /// # Examples
     ///
     /// ```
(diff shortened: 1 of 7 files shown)


src/orderbook/fees.rs
```diff
@@ -109,22 +157,138 @@ impl FeeSchedule {
...
+        // error keeps exactness a checked guarantee rather than a comment.
+        let product = notional
+            .checked_mul(u128::from(bps.unsigned_abs()))
+            .ok_or_else(|| FeeOverflow::new(notional, bps))?;
+        let magnitude = i128::try_from(product / BPS_DENOMINATOR)
+            .map_err(|_| FeeOverflow::new(notional, bps))?;
+        Ok(if bps < 0 { -magnitude } else { magnitude })
+    }
+
+    /// Largest notional whose fee at `bps` is exact (never saturates)
```
joaquinbejar: 🟠 **Warning**: This is the largest notional whose intermediate product fits in `u128`, not the largest notional whose final fee is exact. For `bps = 2`, `u128::MAX / 2 + 1` overflows the multiplication, yet `floor(notional × 2 / 10_000)` still equals the legacy saturated result while `try_calculate_fee` returns `Err`, which also disproves the ‘exact if and only if’ claims above. Either rename and document this as a multiplication-safe bound, or compute the divided result without overflowing and reserve `FeeOverflow` for an unrepresentable final `i128` fee.

joaquinbejar (commented):
The implementation preserves the legacy fee path and respects the module and public API boundaries, but the published exactness contract is stronger than the implementation guarantees. The bound and related documentation need to describe intermediate multiplication safety or the fallible calculation needs to support every representable exact result. REQUEST_CHANGES.

src/orderbook/fees.rs
joaquinbejar: Good catch, the only-if direction was indeed false. Went with the rename-and-document route in 1dd420b, the bound is now max_guaranteed_exact_notional_for_bps documented as the multiplication-safe guarantee, and your 2^127 at 2 bps coincidence is pinned in a test. Kept the conservative rejection rather than widening the arithmetic because the issue contract is rejecting instead of journaling a clamped fee, so Ok has to keep implying the journaled calculate_fee value is exact.

joaquinbejar:
Thanks for the review. Addressed in 1dd420b, the bound is renamed to max_guaranteed_exact_notional_for_bps / max_guaranteed_exact_notional and documented as the multiplication-safe guarantee rather than a tight exactness frontier, with the hairline coincidence case pinned by a test. Kept try_calculate_fee conservative on purpose so Ok always matches the fee calculate_fee would journal.

Issue #199, opened 2026-07-12
Trade-ID namespace is not injectable: Uuid::new_v4() hardcoded in every constructor breaks live/replay ID stability


joaquinbejar:
As of `92db5927ac59` (0.10.4), every `OrderBook` constructor creates its trade/transaction-ID namespace internally with `Uuid::new_v4()`:

```rust
// src/orderbook/book.rs:421-431 (with_clock; new() delegates here)
let namespace = Uuid::new_v4();
...
transaction_id_generator: UuidGenerator::new(namespace),
```

The same pattern repeats in `with_trade_listener` (src/orderbook/book.rs:746) and `with_trade_and_price_level_listener` (src/orderbook/book.rs:794).

`pricelevel::UuidGenerator` is already fully deterministic *given* its namespace — UUID v5 over `namespace + atomic counter` — so the namespace is the only entropy in the trade-ID stream. Because it cannot be injected, trade IDs differ between a live run and its replay even when the command stream, the clock (`with_clock`), and the matching are identical. Consumers building a determinism oracle must exclude or remap upstream trade IDs.

## Suggested work

A pre-publication setter, symmetric with `set_clock`:

```rust
/// Replace the transaction-ID namespace. Call before any orders are
/// submitted (the generator's counter restarts at 0).
pub fn set_trade_id_namespace(&mut self, namespace: Uuid) {
    self.transaction_id_generator = UuidGenerator::new(namespace);
}
```

A `&mut self` setter composes with every existing constructor (`new`, `with_clock`, `with_trade_listener`, …) without multiplying constructor variants. Optionally, a combined `with_clock_and_namespace(symbol, clock, namespace)` convenience constructor for the common deterministic-venue setup.

No wire-format or snapshot change: the namespace is construction state, and a deterministically chosen namespace (e.g. UUID v5 of the symbol under a venue root namespace) yields byte-identical trade IDs across live/replay.

## Context

[...]

PR #202, opened 2026-07-13
Carry the trade-ID namespace in ReplayBookConfig for byte-identical replay (#200)

### LINKED ISSUE #200: Wire the injectable trade-ID namespace into ReplayBookConfig / ReplayEngine (sequencer/replay)
## Overview

#199 made the trade-ID namespace injectable on `OrderBook` (`set_trade_id_namespace`, `with_clock_and_namespace`), but the sequencer's replay surface never uses the new seam: every `ReplayEngine::replay_from*` entry point constructs its book internally via `OrderBook::new` / `OrderBook::with_clock`, which mint a fresh random `Uuid::new_v4()` namespace. Trade IDs produced through the shipped replay API therefore remain non-reproducible — both against the original live run and across repeated replays of the same journal — and the byte-identical-trade-ID guarantee is currently only reachable by hand-rolling the apply loop against a book built with `with_clock_and_namespace`.

## Context

- `src/orderbook/sequencer/replay.rs:281` (`replay_from_with_progress`), `:331` (`replay_from_with_config`), `:416` (`replay_from_with_clock_and_progress`), `:464` (`replay_from_with_clock_and_config`) — all build the book internally with a random namespace.
- `ReplayBookConfig` (`src/orderbook/sequencer/replay.rs:47-72`) carries `fee_schedule`, `stp_mode`, `tick_size`, `lot_size`, `min_order_size`, `max_order_size` — no namespace field; `apply_to` (`:117-131`) is the natural place to apply one.
- Found by the determinism audit of the #199 implementation branch. Companion of the downstream deterministic-venue seams in Option-Chain-OrderBook#147.
- `snapshots_match` is unaffected either way (snapshots never carried the generator), so this is about trade-ID stream reproducibility, not the existing replay-equality oracle.

## Tasks

- [ ] Add `trade_id_namespace: Option<Uuid>` to `ReplayBookConfig` (serde: optional, defaulting to `None` so recorded fixtures keep parsing) and apply it in `apply_to` via `OrderBook::set_trade_id_namespace` before any command is replayed.
- [ ] Decide/document behavior for the non-config entry points (`replay_from_with_progress`, `replay_from_with_clock_and_progress`): keep random namespace (document it) or add namespace-taking variants.
- [ ] Tests: journal a live book built with `with_clock_and_namespace`, replay via `ReplayEngine` with the same namespace + `StubClock`, assert the replayed trade-ID stream is byte-identical to the live one; plus a repeated-replay-same-journal identity test.
- [ ] Docs: `lib.rs` What's New, `CHANGELOG.md`, `make readme`.

## Acceptance criteria

- [ ] A journal replayed through `ReplayEngine` with an injected namespace and clock reproduces the live run's trade IDs byte-identically (test-pinned)
- [ ] `ReplayBookConfig` round-trips through serde with and without the new field (backward-compatible with existing recorded configs)
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `cargo test --all-features` clean
- [ ] `cargo build --release` clean
- [ ] Relevant docs updated (`src/lib.rs` docs, `README.md` regenerated via `make readme`, `CHANGELOG.md`)
- [ ] `snapshots_match` oracle exercised on the new path

## References

- #199 (injectable namespace primitive)
- https://github.com/joaquinbejar/Option-Chain-OrderBook/issues/147 (deterministic venue seams)

joaquinbejar:
## Summary

v0.10.5 (#199) made the trade-ID namespace injectable on `OrderBook`, but the sequencer's replay surface never used the seam: every `ReplayEngine::replay_from*` entry point constructed its book internally with a random `Uuid::new_v4()` namespace, so trade IDs produced through the shipped replay API remained non-reproducible — against the live run and across repeated replays of the same journal. This PR wires the namespace into `ReplayBookConfig`, completing the determinism contract found by the #199 determinism audit.

## Changes

- `ReplayBookConfig.trade_id_namespace: Option<Uuid>` — applied in `apply_to` via `OrderBook::set_trade_id_namespace` before any journal events are replayed. The book is fresh at that point (no orders), so the counter-restart contract of #199 is satisfied by construction.
- `ReplayBookConfig::with_trade_id_namespace(namespace)` — builder-style companion; `ReplayBookConfig::new` keeps its six structural parameters (namespace defaults to `None`), so existing `new(...)` callers are untouched.
- With the live namespace in the config, `replay_from_with_clock_and_config` under an injected `Clock` now reproduces the live trade-ID stream byte-identically. Docs updated across the four entry points; the non-config variants intentionally keep the random namespace (documented) rather than growing more constructor combinations.

## Technical decisions

- **Breaking change, 0.11.0**: `ReplayBookConfig` is a constructible all-pub struct, so the new field breaks exhaustive struct literals (`constructible_struct_adds_field`); the CI `cargo-semver-checks` job enforces this against the crates.io baseline, hence the pre-1.0 breaking bump `0.10.5 → 0.11.0`. CHANGELOG carries the migration note (`trade_id_namespace: None` or `..Default::default()`).
- **No serde work**: the issue's serde round-trip criterion was vacuous — `ReplayBookConfig` derives only `Debug, Clone, Default` and has never had a serde surface; the config is caller-supplied, never read from the journal. No journal or snapshot format change, no `ORDERBOOK_SNAPSHOT_FORMAT_VERSION` bump.
- **Probe-based test oracle**: the generator exposes no ID read-back, so tests prove stream identity via one post-replay crossing pair — trade IDs are UUID v5 of (namespace, counter), so an equal *next* trade ID proves equal namespace and counter position, which proves every prior replayed ID matched the live stream.

## Public API impact

- New: `ReplayBookConfig.trade_id_namespace` field and `ReplayBookConfig::with_trade_id_namespace` method (struct already re-exported at the root and in the prelude — no export changes).
- **Breaking**: exhaustive `ReplayBookConfig { ... }` literals must add the field or use `..Default::default()`.
- `src/lib.rs` What's New 0.11.0; `README.md` regenerated via `make readme`; `CHANGELOG.md` with Added + Breaking sections; version bumped to 0.11.0.

## Testing

- [x] Unit tests in `replay.rs`: builder sets the field / `Default` leaves it `None`; namespace-carrying config reproduces the trade-ID stream of a `with_clock_and_namespace` reference book; default config keeps per-replay random namespaces (`assert_ne` across two replays).
- [x] Integration tests in `tests/unit/replay_config_tests.rs`: end-to-end #200 contract (live book with injected clock + namespace, journaled adds + sweeping market order, replay via `replay_from_with_clock_and_config` → `snapshots_match` holds and probe IDs match live); repeated-replay identity.
- [x] `make pre-push` run locally and clean (1356 tests passed, 0 failed; clippy `-D warnings` clean; release build clean)
- [x] determinism-auditor: zero findings — apply_to ordering verified, no residual entropy in the replay path, probe methodology sound. architect: OK across all areas, semver reasoning confirmed.

## Checklist

- [x] All new `pub` items have `///` documentation
- [x] Module boundaries respected (sequencer → book direction only; core engine untouched)
- [x] No new dependencies added
- [x] `src/lib.rs` docs + `README.md` (via `make readme`) + `CHANGELOG.md` updated

Closes #200

### DIFF
--- a/src/orderbook/sequencer/replay.rs
+++ b/src/orderbook/sequencer/replay.rs
@@ -16,6 +16,7 @@ use serde::{Deserialize, Serialize};
 use std::marker::PhantomData;
 use std::sync::Arc;
 use thiserror::Error;
+use uuid::Uuid;
 
 /// Book configuration injected into a fresh [`OrderBook`] before replay so
 /// that a journal produced by a **non-default-config** book reconstructs to
@@ -40,6 +41,13 @@ use thiserror::Error;
 /// The configuration is supplied by the **caller** — replay does not read it
 /// from the journal, so the on-disk journal format is unchanged and
 /// `ORDERBOOK_SNAPSHOT_FORMAT_VERSION` is not bumped.
+///
+/// Beyond the structural fields, the config can also carry the source book's
+/// **trade-ID namespace** (`trade_id_namespace`, see
+/// [`OrderBook::set_trade_id_namespace`]): with it, a `*_with_config` replay
+/// under an injected [`Clock`] reproduces the live run's trade-ID stream
+/// byte-identically, not just its structure. Without it the fresh book keeps
+/// a random namespace and replayed trade IDs differ from the live ones.
 ///
 /// `Default` yields the all-defaults configuration, equivalent to the plain
 /// replay entry points.
@@ -69,14 +77,25 @@ pub struct ReplayBookConfig {
     /// Maximum order size the source book used, or `None` for no maximum.
     /// Applied via [`OrderBook::set_max_order_size`] only when `Some`.
     pub max_order_size: Option<u64>,
+
+    /// Trade-ID namespace the source book used, or `None` to keep the fresh
+    /// book's random namespace. Applied via
+    /// [`OrderBook::set_trade_id_namespace`] only when `Some`, before any
+    /// journal events are replayed (the fresh book has no orders, so the
+    /// generator's counter-restart contract is satisfied by construction).
+    /// Inject the live book's namespace (#199) to make the replayed trade-ID
+    /// stream byte-identical to the original.
+    pub trade_id_namespace: Option<Uuid>,
 }
 
 impl ReplayBookConfig {
-    /// Creates a [`ReplayBookConfig`] from its six fields.
+    /// Creates a [`ReplayBookConfig`] from its six structural fields.
     ///
     /// Equivalent to building the struct with public-field syntax; provided so
     /// callers can construct the carrier without naming every field at the call
-    /// site. Use [`ReplayBookConfig::default`] for the all-defaults case.
+    /// site. Use [`ReplayBookConfig::default`] for the all-defaults case. The
+    /// `trade_id_namespace` field defaults to `None` — chain
+    /// [`Self::with_trade_id_namespace`] to set it.
     ///
     /// # Arguments
     ///
@@ -102,7 +121,24 @@ impl ReplayBookConfig {
             lot_size,
             min_order_size,
             max_order_size,
+            trade_id_namespace: None,
         }
+    }
+
+    /// Returns this configuration with the trade-ID namespace set.
+    ///
+    /// Builder-style companion to [`Self::new`] (which leaves the namespace
+    /// at `None`): carry the live book's namespace (#199) so a
+    /// `*_with_config` replay under an injected [`Clock`] reproduces the
+    /// live trade-ID stream byte-identically.
+    ///
+    /// # Arguments
+    ///
+    /// * `namespace` — trade-ID namespace the source book used
+    #[must_use = "with_trade_id_namespace returns the updated config; it does not mutate in place"]
+    pub fn with_trade_id_namespace(mut self, namespace: Uuid) -> Self {
+        self.trade_id_namespace = Some(namespace);
+        self
     }
 
     /// Applies this configuration to a freshly-constructed `book` in place,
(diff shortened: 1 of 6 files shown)


src/orderbook/sequencer/replay.rs
```diff
@@ -128,6 +167,9 @@ impl ReplayBookConfig {
         if let Some(max) = self.max_order_size {
             book.set_max_order_size(max);
         }
+        if let Some(namespace) = self.trade_id_namespace {
+            book.set_trade_id_namespace(namespace);
```
joaquinbejar, line 185: 🔴 **Critical**: This always restarts the UUID counter at zero, but the config replay entry points still accept an arbitrary `from_sequence` and build a fresh book. A self-contained journal suffix after earlier live trades will therefore mint the wrong IDs and can reissue IDs already emitted under this namespace. Carry the generator's starting counter in the replay config, or otherwise require and validate that replay begins at the trade-ID stream origin before claiming byte-identical IDs.

joaquinbejar (commented):
The namespace is wired into the replay path cleanly, but resetting its generator to zero is not sufficient for suffix replay and can reproduce wrong or duplicate trade IDs after earlier live trades. REQUEST_CHANGES.

src/orderbook/sequencer/replay.rs
joaquinbejar: Right, the counter restart made suffix replay actively dangerous, not just undocumented. Fixed in 42fe894 with a typed guard, both config entry points now reject a namespace-carrying config when from_sequence != 0 via the new ReplayError::NamespaceRequiresFullReplay, while namespace-free suffix replay keeps working. I went with validation over carrying a starting counter because seeding the generator needs an upstream pricelevel::UuidGenerator API and the engine can't derive the seq-to-trade-count mapping anyway, and the remaining rotated-journal caveat is documented on the field.

joaquinbejar:
Good catch on the suffix-replay hole. Addressed in 42fe894, namespace injection now requires a full replay (typed NamespaceRequiresFullReplay error, InvalidSequence keeps precedence, namespace-free suffix replay unchanged), and the rotated-journal caveat the engine can't detect is documented on the field. Carrying a starting counter would need an upstream UuidGenerator API, happy to file that in pricelevel if suffix replay with reproducible IDs ever becomes a real need.

joaquinbejar (commented):
The namespace guard now prevents unsafe suffix replay in both configurable entry points, preserves the existing namespace-free path and error precedence, and documents the remaining rotated-journal caveat clearly. APPROVE.

PR #204, opened 2026-07-14
Document and property-test the queue-priority contract (#203)

### LINKED ISSUE #203: Review request: external Tracebook conformance adapter
Hi, I maintain [Tracebook](https://github.com/Taz33m/tracebook), an external semantic-conformance and failure-reduction toolkit for matching engines. Release 0.4.0 includes a native process adapter around `orderbook-rs` 0.10.4:

- [adapter source](https://github.com/Taz33m/tracebook/tree/v0.4.0/integrations/orderbook_rs)
- [0.4.0 release and five-event proof](https://github.com/Taz33m/tracebook/releases/tag/v0.4.0)
- [`fifo-limit-v1` capability contract](https://github.com/Taz33m/tracebook/blob/v0.4.0/docs/conformance.md#fifo-limit-v1-capability-profile)

The correct adapter passes our deterministic FIFO lifecycle cases and a 1,000-event generated `fifo-full-v1` campaign. It runs as an independent Rust process; no Tracebook reference-matching code is used candidate-side.

Could a maintainer sanity-check whether these adapter assumptions match the intended public semantics?

1. Quantity reduction preserves an order's FIFO priority.
2. Replacement is modeled as remove plus add and therefore loses FIFO priority.
3. Resting queue order and maker/taker IDs are reconstructed faithfully through the public APIs we use.
4. Since upstream is now at 0.11.0, whether you would prefer us to update the reproducible 0.10.4 pin before treating this adapter as a useful external example.

The release also contains a separately named `faulty-orderbook-adapter`. Its stale-priority behavior is deliberately injected in Tracebook's adapter state for demonstration; it is **not** a claim of an `orderbook-rs` bug. The demo detects the injected mismatch at event 173 and reduces it to five events.

No upstream code change is requested. I am looking for semantic review, and I will update the adapter/docs based on any corrections.

joaquinbejar:
## Summary

Issue #203 (Tracebook, an external conformance toolkit) revealed that the queue-priority semantics of `OrderBook::update_order` are consumed as a public contract, yet they were documented only inside the embedded `pricelevel` engine and had no regression coverage in this crate. This PR promotes the contract to the public docs and locks it in with property-based tests. No behavior change.

## Changes

- Documented the queue-priority contract on `OrderBook::update_order`: an in-place quantity decrease (or unchanged total) keeps the maker's queue position; a quantity increase demotes the order to the back of its price level; `UpdatePrice` / `UpdatePriceAndQuantity` / `Replace` are cancel-then-add and always lose time priority — for `Replace` and `UpdatePriceAndQuantity` even when the price is unchanged.
- Documented the snapshot caveat: a demoted order keeps its original admission timestamp (only its insertion sequence is refreshed), so `(timestamp, seq)`-sorted snapshot views can show it ahead of where matching will consume it.
- Added `tests/unit/props_quantity_update_priority.rs` with two proptest invariants (256 cases each) driven purely through the public API: a market-order sweep after an in-place decrease must fill the resized order at its original queue position, and a sweep after an increase must fill it last.
- `CHANGELOG.md` entry under Unreleased.

## Testing

- [x] Unit tests added or updated
- [x] Integration tests under `tests/unit/` added or updated
- [x] Pre-submission checks run locally and clean (`cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, `cargo test --all-features`, `cargo build --release`)

## Checklist

- [x] Module boundaries respected (core engine has no deps on `manager`, `nats*`, `sequencer/`)
- [x] No new dependencies added without explicit approval
- [x] `CHANGELOG.md` updated (`src/lib.rs` module docs untouched, so no `README.md` regeneration needed)

Closes #203

### DIFF
--- a/src/orderbook/modifications.rs
+++ b/src/orderbook/modifications.rs
@@ -122,6 +122,28 @@ where
 {
     /// Update an order's price and/or quantity
     ///
+    /// # Queue priority
+    ///
+    /// The update variants follow conventional exchange price-time-priority
+    /// rules. This is a public contract — external conformance tooling
+    /// depends on it (see issue #203):
+    ///
+    /// - [`OrderUpdate::UpdateQuantity`] with a **decreased or unchanged**
+    ///   total quantity (visible + hidden) updates the resting order in
+    ///   place at its existing insertion sequence: the maker keeps its
+    ///   queue position. Reducing size never forfeits time priority.
+    /// - [`OrderUpdate::UpdateQuantity`] with an **increased** total
+    ///   quantity demotes the order to the back of its price level's
+    ///   queue. Sizing up loses time priority. The demoted order keeps
+    ///   its original admission timestamp — only its insertion sequence
+    ///   is refreshed — so snapshot views sorted by `(timestamp, seq)`
+    ///   may show it ahead of where matching will actually consume it.
+    /// - [`OrderUpdate::UpdatePrice`], [`OrderUpdate::UpdatePriceAndQuantity`],
+    ///   and [`OrderUpdate::Replace`] are implemented as cancel-then-add:
+    ///   the order always re-enters at the back of its (possibly new)
+    ///   price level and loses time priority — for `Replace` and
+    ///   `UpdatePriceAndQuantity` even when the price is unchanged.
+    ///
     /// # Errors
     /// Returns [`OrderBookError::KillSwitchActive`] when the kill switch
     /// is engaged and the update is anything other than
--- a/src/orderbook/modifications.rs
+++ b/src/orderbook/modifications.rs
@@ -136,8 +136,13 @@ where
     ///   quantity demotes the order to the back of its price level's
     ///   queue. Sizing up loses time priority. The demoted order keeps
     ///   its original admission timestamp — only its insertion sequence
-    ///   is refreshed — so snapshot views sorted by `(timestamp, seq)`
-    ///   may show it ahead of where matching will actually consume it.
+    ///   is refreshed. **Known limitation:** the demotion does not yet
+    ///   survive a snapshot round-trip — level capture orders by
+    ///   `(timestamp, seq)` without persisting the sequence, and
+    ///   [`restore_from_snapshot`](OrderBook::restore_from_snapshot)
+    ///   re-enqueues in that order, so a restored book consumes the
+    ///   upsized order at its pre-demotion position (tracked in
+    ///   issue #205, fix upstream in `pricelevel`).
     /// - [`OrderUpdate::UpdatePrice`], [`OrderUpdate::UpdatePriceAndQuantity`],
     ///   and [`OrderUpdate::Replace`] are implemented as cancel-then-add:
     ///   the order always re-enters at the back of its (possibly new)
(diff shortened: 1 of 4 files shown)


src/orderbook/modifications.rs
```diff
@@ -122,6 +122,28 @@ where
...
+    /// - [`OrderUpdate::UpdateQuantity`] with a **decreased or unchanged**
+    ///   total quantity (visible + hidden) updates the resting order in
+    ///   place at its existing insertion sequence: the maker keeps its
+    ///   queue position. Reducing size never forfeits time priority.
+    /// - [`OrderUpdate::UpdateQuantity`] with an **increased** total
+    ///   quantity demotes the order to the back of its price level's
+    ///   queue. Sizing up loses time priority. The demoted order keeps
+    ///   its original admission timestamp — only its insertion sequence
+    ///   is refreshed — so snapshot views sorted by `(timestamp, seq)`
+    ///   may show it ahead of where matching will actually consume it.
```
joaquinbejar: 🔴 **Critical**: This understates the impact. The snapshot persists this `(timestamp, seq)` order, and `PriceLevel::from_snapshot` rebuilds a fresh queue by pushing that vector in order, so an upsized older order can regain front priority after restore. That changes maker and trade order after recovery, violating both this public priority contract and the snapshot round-trip requirement. Preserve insertion order through snapshot and restore, and add an upsize then restore then sweep regression test.

tests/unit/props_quantity_update_priority.rs
```diff
@@ -0,0 +1,185 @@
...
+//! `OrderBook::update_order(OrderUpdate::UpdateQuantity { .. })` (issue #203):
+//!
+//! - A quantity **decrease** (or unchanged total) keeps the maker's queue
+//!   position — reducing size never forfeits time priority.
+//! - A quantity **increase** demotes the order to the back of its price
+//!   level's queue.
+//!
+//! Both properties are asserted through the public API only: build a level of
+//! resting standard orders, resize one, then sweep the level with an
+//! aggressive market order and check the maker-id sequence of the fills.
```
joaquinbejar, line 14: 🟠 **Warning**: These properties cover only `UpdateQuantity`, while the contract added in this PR also promises that same-price `Replace` and `UpdatePriceAndQuantity` always lose priority. Add public-API sweep assertions for those two cases so the newly promoted replacement contract cannot regress without detection.

joaquinbejar (commented):
The quantity-update tests are solid, but snapshot restore can currently undo the documented upsize demotion, and the newly promoted replacement guarantees are not covered by maker-order assertions. REQUEST_CHANGES.

src/orderbook/modifications.rs
joaquinbejar: Good catch, the restore path really does rebuild the queue from the (timestamp, seq) vector and undoes the demotion. The fix belongs in pricelevel's snapshot ordering so I tracked it upstream in joaquinbejar/PriceLevel#109 plus #205 here, and the doc block now states the restore limitation explicitly instead of understating it.

tests/unit/props_quantity_update_priority.rs
joaquinbejar: Agreed, added property tests for same-price Replace and UpdatePriceAndQuantity that sweep the level and assert the updated order always fills last, regardless of quantity direction. Pushed in e5464f8.

joaquinbejar:
Thanks for the review. Same-price Replace and UpdatePriceAndQuantity coverage is in (e5464f8), and the snapshot-restore regression is confirmed and tracked in #205 with the upstream ordering fix in joaquinbejar/PriceLevel#109, the doc block now states that limitation explicitly.

Issue #168, opened 2026-06-24
fix(modifications): extend modify atomicity to STP self-cross taker-cancellation


joaquinbejar:
## Overview

> Created from the #98 (validate-first atomic modify) microstructure review.

#98 made `UpdatePrice` / `UpdatePriceAndQuantity` / `Replace` validate-first for
every **pre-match admission** rejection (risk, STP missing user_id, tick, lot,
min/max size, expiry, post-only-would-cross, FOK). One **post-match** rejection
remains uncovered: if a re-price/replace moves the order to a price that crosses
into the **same user's** resting liquidity on the opposite side and STP is in
`CancelTaker` or `CancelBoth` mode, `add_order` matches post-cancel and returns
`SelfTradePrevented` (the taker is cancelled) — **after** the original was already
removed. The original order is destroyed.

This is narrower than the #98 paths (requires the user to re-price their own order
through their own opposite-side resting depth under CancelTaker/CancelBoth), and
it was not in #98's enumerated reject list, so it was deliberately deferred.

## Proposed approach

Add an STP self-cross pre-check to the modify guard (before `cancel_order`):
walk the opposite side for same-user resting makers in the crossable range (the
machinery already exists — `fok_fillable_quantity` walks the opposite side with
`taker_user_id` and STP awareness; `check_stp_at_level` computes the per-level STP
action). If, under the active STP mode, the new order's taker would be cancelled
(`CancelTaker` / `CancelBoth`), reject the modify pre-cancel with
`SelfTradePrevented` so the original survives. `CancelMaker` does not cancel the
taker (the new order would rest), so it needs no pre-check.

Add a test: a user with a resting order on each side, modifying one to cross their
own opposite-side order under `CancelTaker` — assert the original survives and the
typed STP error is returned.

## Acceptance criteria

- [ ] A re-price/replace that would self-cross under CancelTaker/CancelBoth is rejected before cancel; the original survives unchanged
- [ ] CancelMaker behavior unchanged (no spurious pre-rejection)
- [ ] Test covers the self-cross-on-modify case
- [ ] Standard gates clean (clippy/fmt/test/release)

## References
- `src/orderbook/modifications.rs` (`validate_order_shape`, the three modify branches)
- `src/orderbook/matching.rs` (`fok_fillable_quantity`, STP taker-cancel arms)
- `src/orderbook/stp.rs` (`check_stp_at_level`)

Issue #124, opened 2026-06-23
fix(modifications): replace redundant Arc::try_unwrap(order.clone()) with a direct deep clone in modify


joaquinbejar:
## Overview

In UpdatePrice and UpdatePriceAndQuantity, Arc::try_unwrap(order.clone()).unwrap_or_else(|arc| (*arc).clone()) clones the Arc (bumping the refcount) then tries to unwrap it; because the clone keeps strong count >= 2, try_unwrap always fails and falls through to a full deep clone. The cheap branch is dead and the intermediate Arc clone is pure overhead.

**Why this matters.** Misleading code that reads as an optimization but never takes the cheap branch, on the modify path. Behavior is already correct; this is a clarity/minor-overhead cleanup. order is not used after this point.

> Surfaced by an automated multi-agent code review (adversarially verified). Severity: **low** · Effort: **S** · Category: **refactor**.

## Context

Affected files:

- `src/orderbook/modifications.rs`

See `rules/global_rules.md` and `CLAUDE.md` for the binding conventions.

## Proposed approach

Replace both occurrences with a direct deep clone of the underlying value: let original_order = (*order).clone();. Verify the modify tests still pass.

## Acceptance criteria

- [ ] Both modify branches use a direct (*order).clone() with no redundant Arc clone + try_unwrap
- [ ] Modify behavior is unchanged and existing modify tests pass
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `cargo test --all-features` clean
- [ ] `cargo build --release` clean (zero warnings)
- [ ] Docs / `CHANGELOG.md` updated as needed (`make readme` if public surface changed)

PR #222, opened 2026-09-09
Trigger self-trade prevention only on a reachable same-user maker

Swofty-Developments:
## Summary

`check_stp_at_level` reports a self-trade conflict whenever a same-user maker rests at a crossed level, whether or not the taker can actually reach it, and the `CancelTaker` / `CancelBoth` arms then cancelled the taker (and, under `CancelBoth`, the maker) unconditionally after their safe-quantity pre-match. Ask 5 @100 (user 2) queued ahead of ask 9 @100 (user 1), buy 3 @100 as user 1: the taker fills entirely against user 2, never reaches its own maker, and still returned `SelfTradePrevented` with `Cancelled { filled_quantity: 3 }`, so a client that retried double-filled. Under `CancelBoth` the untouched 9-lot maker was cancelled with it, and market takers reach the same arms through `match_order_with_user`, which drops the taker flag, so the caller saw `Ok` while the maker was gone.

## Changes

- **The two taker-cancelling arms cancel only on a reachable maker.** `CancelTaker` and `CancelBoth` now set `stp_taker_cancelled` (and, for `CancelBoth`, cancel the maker) only when the sweep can still execute into the same-user maker. Two guards let the sweep past an unreachable one: a budget the pre-match exhausted (`StopCondition::is_done`), which is an ordinary complete fill, and a quote-notional residual that cannot fund one lot at this level's price (`StopCondition::is_dust_at`), which walks on to the next level rather than breaking, because a notional sell can still afford a whole lot at a cheaper bid.
- **The dust guard is quote-only.** A base-quantity residual keeps the STP verdict whatever its size: a maker admitted before a `set_lot_size` change keeps resting with a misaligned tranche, so a sub-lot residual is genuinely reachable, and walking past it would rest the taker crossed against its own maker.
- **The modify precheck agrees with the sweep.** `check_modify_stp_self_cross` (#168) rejected a reprice as soon as any same-user order rested at a crossed level, before subtracting the non-self depth queued ahead of it. It now consults the same insertion-sequence `check_stp_at_level` verdict and mirrors the sweep's lot-rounded per-level cap, so a reprice the non-self depth covers is admitted, one it does not cover is still refused before the original is cancelled, and dust the sweep stops on before a deeper same-user level is admitted exactly as a direct submit is.
- **The modify precheck is bounded by the level's own dry run.** The `CancelTaker` / `CancelBoth` arm of `check_modify_stp_self_cross` subtracted the visible depth `check_stp_at_level` counted ahead of the same-user maker, while the sweep subtracts what `PriceLevel::match_order` executes. A counted maker can deliver less, one that makes no progress is set aside and a replenish whose checked net delta would overflow the level's visible counter aborts the sweep untouched (#124), so the precheck admitted a reprice the sweep then killed after the original was cancelled. It now bounds the pre-match with `PriceLevel::matchable_quantity`, the authoritative dry run the no-conflict arm already used.
- **A quote-notional sell no longer stops at an unaffordable level.** The per-level guard ended the whole walk on a zero `level_qty_cap`. That is right for a base-quantity budget (the cap is the lot-rounded residual, price-independent) and for a notional buy (asks ascend, so `remaining / price` only shrinks), and wrong for a notional sell, whose walk descends the bids so a budget too small here can still fund a lot lower down: a sell of 150 into bids 100, 75 and 50 executed one unit instead of two. The direction-aware decision now lives in `StopCondition::zero_cap_is_terminal(taker_side)` next to `is_dust_at`, documented with the price-monotonicity argument, and the walk skips the level instead of breaking when a notional sell can still reach cheaper bids. Prices are at least one, so the cap at any level is at most the remaining notional and `remaining < lot` is an exact terminal test, which keeps a dust budget from scanning a deep book.

## Technical decisions

- `fok_fillable_quantity` needs no mirror. Its `(safe_quantity, stop = true)` arm already reports `min(cap, safe_quantity)`, so a taker covered by the non-self depth was always feasible; the divergence was the sweep, which reported a kill after executing. The two now agree.
- Market orders are fixed at the source rather than by threading `MatchOutcome.taker_stp_cancelled` out to the market path.
- `zero_cap_is_terminal` is consulted only once the cap is already zero, so the base-quantity hot path evaluates nothing new. Its sell arm means a small notional budget can visit every remaining bid level, bounded by the level count the sweep would traverse anyway.
- A taker that fills entirely from non-self depth now returns `Ok`; a reachable maker still yields `SelfTradePrevented` with the true non-self fill.
- Scope: the reachability rule covers `STPMode::CancelTaker` and `STPMode::CancelBoth`. `CancelMaker` is deliberately unchanged and still cancels every same-user order at a level the sweep touches. `STPMode` gains a `# Reachability` section recording that, plus the asymmetry lot rounding leaves: a sub-lot residual at the conflicting level cancels the taker, while the identical residual arising one level earlier rests crossed against the same user's maker, because the top-of-loop zero-cap check precedes the STP scan.

## Testing

`tests/unit/stp_reachability_tests.rs`, 19 tests, each deriving its expected numbers level by level: the original repro under `CancelTaker` and `CancelBoth` and as fill-or-kill and market taker; the exact-depth boundary; quote-amount buys, both dust and reachable; a quote-amount sell past an unaffordable same-user level; a lot-rounded notional residual; a sub-lot reserve tranche resting from before a `set_lot_size` change, on the sweep and on a reprice; precheck and sweep agreement on dust before a deeper same-user level; same-level reprices, admitted and refused; and the new intermediate-level regressions, a notional sell walking past an unaffordable bid to a cheaper one under both modes, a notional buy still stopping at the first unaffordable ask, and a base-quantity sub-lot residual still ending the walk. A precheck regression drives the #124 replenish-headroom abort through the public API and asserts the reprice is refused with the original still resting, where before the fix it destroyed the original. Six unit tests cover `zero_cap_is_terminal` directly, including the exact `remaining < lot` boundary. Two non-STP cases in `tests/unit/market_order_by_amount_tests.rs` pin the walk on a plain book: the sell that skips 75 to fill at 50, and a lot-10 book where no cheaper bid can fund a lot and both are left untouched.

Full suite on this head: 870 lib + 581 integration + 26 fee + 6 metrics + 1 alloc-budget + 49 doc, 0 failed; `make test` 737 + 543 + 26 + 46, 0 failed. `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo build --release` clean.

`src/lib.rs` carries the 0.13.0 What's New entries (README regenerated) and `CHANGELOG.md` an `[Unreleased]` entry describing the reachability rule, the quote-versus-base dust policy, the modify precheck and the level-walk fix. No public API change, no snapshot or journal format change.

Rebased onto `main` after #227, #229, #231 and #232; the maintainer finished the last review item (the quote-sell level walk) and the description on this branch.


src/orderbook/matching.rs
```diff
@@ -588,6 +598,12 @@ where
                                 stop.consume(executed, price);
                             }
                         }
+                        // Same reachability rule as `CancelTaker` above, and
+                        // here it also gates the maker cancellation: a maker
+                        // the taker never reached must survive untouched.
+                        if stop.is_done() {
```
joaquinbejar, line 794: 🔴 **Critical**: This still fires on the notional path. `is_done()` is exact zero on the budget, and a quote amount sweep normally ends with dust below one unit, not zero. Same book as your tests under `CancelBoth`, `submit_market_order_by_amount_with_user` for 350 at the 100 level fills 3 lots, leaves `remaining` at 50, so this arm cancels the untouched 9 lot maker and the caller gets `Ok`. I ran that against this branch and the maker ends up `Cancelled { filled_quantity: 0, reason: SelfTradePrevention }`.

After matching the safe non-self depth, check whether any quantity remains executable at this price before cancelling either order. Apply this in both `CancelTaker` and `CancelBoth`. I have withdrawn the earlier one-line suggestion because an unconditional break on zero `level_qty_cap` can stop quote amount sells before an affordable lower bid. See the follow-up in this thread for that reproduction. Cover the lot-rounded residual behavior explicitly as well.

tests/unit/stp_reachability_tests.rs
```diff
@@ -0,0 +1,249 @@
...
+            .expect("feasible FOK must not be killed after executing");
+
+        assert_eq!(filled(&trades), 5, "FOK filled its complete quantity");
+        assert_self_maker_intact(&book);
+    }
+
+    /// A market taker takes the same path with the STP flag dropped, so
+    /// `CancelBoth` cancelled the maker while returning `Ok`.
+    #[test]
+    fn market_cancel_both_leaves_unreachable_maker_intact() {
```
joaquinbejar, line 214: 🟠 **Warning**: Worth adding the notional twin of this test, `submit_market_order_by_amount_with_user` for 350 under `CancelBoth` with the same seed book. Today it fails on this branch, the maker is cancelled and the call still returns `Ok` with 3 filled. That is the one case the base quantity tests cannot see because their budget lands exactly on zero.

joaquinbejar (dismissed):
The base quantity fix and FOK feasibility check now agree, but the quote amount path still cancels an unreachable maker when the budget ends in dust. I reproduced this on the branch. Copilot's modify concern is also valid. Repricing a buy of 3 from 90 to 100 still returns SelfTradePrevented with 5 non-self units ahead of the same-user maker at 100, so the modify precheck needs the same FIFO reachability rule. Please add quote amount and lot rounding regressions and an [Unreleased] entry to CHANGELOG.md. REQUEST_CHANGES.

src/orderbook/matching.rs
joaquinbejar: 🔴 **Critical**: Correction to my suggested guard. `stop.is_done()` ends the sweep, but `stop.level_qty_cap(price, lot) == 0` only rules out execution at the current price. For a quote amount sell of 150 against other-user 1@100, same-user 1@100, and other-user 1@50, the remaining 50 can still execute at 50. Preserve the self maker and consider the next price level instead of unconditionally breaking.

Apply this distinction to both `CancelTaker` and `CancelBoth`. Please add this sell case and lot rounding coverage alongside the buy-budget regression. The buy reproduction and the missing test reported above remain valid.

Swofty-Developments: I've fixed this with a second guard in my latest commits

src/orderbook/matching.rs
```diff
@@ -550,6 +550,28 @@ where
...
+                        // afford a whole lot at a cheaper bid further down,
+                        // so walk on rather than break (for a buy the next
+                        // ask is dearer and the loop's own cap check ends the
+                        // sweep). `check_modify_stp_self_cross` dry-runs the
+                        // same decision on the modify path (#168).
+                        if stop.is_done() {
+                            break;
+                        }
+                        if stop.level_qty_cap(price, lot) == 0 {
+                            continue;
```
joaquinbejar, line 709: 🟠 P2 — Quote-sell continuation still stops before affordable deeper bids. I ran an additional regression on this head: bids 100:[other-user 1, same-user 1], 75:[other-user 1], 50:[other-user 1], then a same-user quote-amount sell of 150. Expected fills are 1 at 100 and 1 at 50; actual execution is only 1 at 100. This new `continue` advances to 75, where the loop's `qty_cap == 0` check at lines 440–442 breaks and never visits 50. This is an incomplete fix, rather than a newly introduced regression. Make the outer zero-cap handling distinguish an exhausted budget from a quote sell that can still execute at cheaper bids, and cover an intervening unaffordable level under both CancelTaker and CancelBoth.

joaquinbejar (commented):
Re-reviewed the latest head. All 12 new STP regression tests pass; the earlier exact-fill and quote-buy-dust findings are resolved. One additional quote-sell regression still fails (inline below). Copilot's lot-aware modify-precheck mismatch also reproduces, but #231 rejects its freshly admitted Reserve 3/2 fixture; after updating main, assess it specifically for legacy/off-lot resting liquidity (e.g. orders seeded before set_lot_size), not as a newly valid admission. Please also rewrite the PR description to match the latest implementation/tests/CHANGELOG and resolve the current merge conflict before rerunning CI.

src/orderbook/matching.rs
joaquinbejar: Re-tested on latest head 0a95f12: this P2 is still reproducible independently under BOTH CancelTaker and CancelBoth. With bids 100:[other 1, self 1], 75:[other 1], 50:[other 1] and a quote-amount sell of 150, execution remains 1 unit rather than 2; the bid at 50 is untouched. The new quote-only is_dust_at guard fixes the base-quantity distinction, but its continue at matching.rs:592–593 / 678–679 still reaches the unchanged outer cap==0 break at :458–460. Please also skip unaffordable intermediate bids for a non-exhausted quote sell, while retaining the separately intended base-quantity behavior. The 15 shipped reachability tests pass; the two intermediate-level regression probes fail.

joaquinbejar (commented):
Re-reviewed 0a95f12. All 15 shipped STP-reachability regressions pass. The new quote-only dust policy and lot-rounded modify precheck resolve the previously reported same-level legacy sub-lot disagreement; no additional architectural/API issue found.

The existing P2 remains open: quote sells still stop at an intermediate unaffordable bid. I reran the 150-budget / 100→75→50 regression separately under CancelTaker and CancelBoth, and both execute 1 unit instead of 2. I updated the existing thread with the current line references and results: https://github.com/joaquinbejar/OrderBook-rs/pull/222#discussion_r3996310877 . Fix the outer zero-cap handling for quote sells and add both regression cases before merge.

Please also refresh the PR description: it still says the guard is only is_done(), that check_modify_stp_self_cross already had the same rule, and that no CHANGELOG entry was added. Describe the current quote-versus-base dust policy, the modified precheck, and the added tests/release note.

src/orderbook/modifications.rs
```diff
@@ -1202,20 +1212,37 @@ where
...
+                    remaining = remaining
+                        .saturating_sub(level.matchable_quantity(remaining, new_order.id()));
+                }
+                STPAction::CancelTaker { safe_quantity }
+                | STPAction::CancelBoth { safe_quantity, .. } => {
+                    // The sweep fills up to `safe_quantity` against the non-self
+                    // depth queued ahead of the same-user maker and cancels the
+                    // taker only if quantity is still left after that. A taker
+                    // the non-self depth satisfies never reaches its own maker.
+                    if remaining > safe_quantity {
```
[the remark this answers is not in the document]
joaquinbejar: Addressed in 0a95f12. The modify precheck now walks levels with the sweep's lot-rounded cap, so a sub-lot residual stops the dry run exactly where the sweep stops, and the reserve plus lot regressions ship alongside it.

CHANGELOG.md
```diff
@@ -9,6 +9,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Self-trade prevention fires only on a same-user maker the taker can
+  reach (#222).** `check_stp_at_level` reports a conflict whenever a
+  same-user maker rests at a crossed level, and the `CancelTaker` /
+  `CancelBoth` arms then cancelled the taker (and, under `CancelBoth`, the
```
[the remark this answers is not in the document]
joaquinbejar: The changelog hunk is correct, I asked for the Unreleased entry in review, so it is the PR description that is out of date rather than the diff. The author has already been asked to rewrite the description to match the current implementation, tests and release note.

joaquinbejar:
Status on 0a95f12 so the remaining work is clear. The two Copilot threads are closed, the lot-rounded modify precheck landed and the changelog entry is the one I asked for, so only the PR description is stale there. The base quantity versus quote dust distinction is right now that is_dust_at is quote only, and the stale test comment is gone.

One blocker is left, the quote sell that stops at an intermediate unaffordable bid, still reproducible under both CancelTaker and CancelBoth with bids 100 other 1 and self 1, 75 other 1, 50 other 1 and a quote amount sell of 150, executing 1 unit instead of 2. Details and line references are in https://github.com/joaquinbejar/OrderBook-rs/pull/222#discussion_r3996310877 .

Also still open, the merge conflict against main and the PR description rewrite. Main has moved since this branch was cut, #227, #229 and #231 landed and #232 is in review, so rebase before the next CI run.

src/orderbook/matching.rs
joaquinbejar: Fixed on 4bd5866. The zero cap check now asks whether it is terminal for this walk direction, terminal for a base quantity budget and for a notional buy walking up the asks, not terminal for a notional sell walking down the bids, so the sell skips the unaffordable level and reaches the cheaper one. Your 150 into 100, 75 and 50 regression now executes 2 units under both CancelTaker and CancelBoth, and it ships as a test along with a notional buy twin and a base quantity sub lot case that both still stop the walk.

Issue #221, opened 2026-09-08
Reserve UpdatePriceAndQuantity can silently ignore requested quantity increase


mcc0nnell:
OrderBook::update_order(OrderUpdate::UpdatePriceAndQuantity) appears to apply inconsistent quantity semantics for ReserveOrder.

The underlying pricelevel::OrderUpdate::UpdateQuantity contract defines new_quantity as the visible tranche for two-tranche orders, leaving hidden quantity unchanged. OrderBook follows that behavior for UpdateQuantity, and Replace also writes the reserve order’s visible quantity directly.

However, UpdatePriceAndQuantity clones the order and calls OrderQuantity::set_quantity(new_quantity). For ReserveOrder, set_quantity delegates to reduce_reserve_to_total, which interprets the argument as a target total quantity rather than a new visible quantity.

**Minimal reproduction**

Starting order:

ReserveOrder
price   = 1000
visible = 5
hidden  = 5
total   = 10

Modify:

OrderUpdate::UpdatePriceAndQuantity {
    order_id: id,
    new_price: Price::new(1010),
    new_quantity: Quantity::new(15),
}

Expected from the documented visible-tranche semantics:

price   = 1010
visible = 15
hidden  = 5
total   = 20

The current ReserveOrder path through reduce_reserve_to_total computes:

10.saturating_sub(15) == 0

so the resulting quantity remains effectively:

price   = 1010
visible = 5
hidden  = 5
total   = 10

The operation returns success and changes the price, while the requested quantity increase is not applied.

A more amplified case is 30 visible / 70 hidden with new_quantity = 80: the visible-tranche interpretation yields 80/70, while the total-target interpretation reduces the visible tranche to 10/70.

**Existing test gap**

test_update_order_with_all_types already constructs a reserve order with visible=5, hidden=5, performs UpdatePriceAndQuantity(... new_quantity=15), and checks that the operation succeeds and that the price changes. It does not assert the resulting reserve quantity, so this divergence can pass the suite.

**Likely fix**

Keep UpdatePriceAndQuantity aligned with UpdateQuantity and the pricelevel contract for two-tranche orders: treat new_quantity as the new visible tranche while preserving hidden quantity.

Reserve total-remainder adjustment should remain confined to the separate residual-resting path (set_total_remaining).

I have a minimal regression fixture/test ready if useful.

PR #223, opened 2026-09-09
Cancel the order on an UpdateQuantity to zero

Swofty-Developments:
## Summary

`OrderUpdate::UpdateQuantity { new_quantity: 0 }` was accepted and applied as a resize. pricelevel keeps the maker in its queue (the projected total is below the live total, so it is a keep-in-place resize), leaving an order that can never fill resting at zero depth. It held `best_bid` / `best_ask` on a level with nothing behind it, so `will_cross_market` rejected a post-only at that price against liquidity that did not exist, and the sweep that eventually dropped it emitted no trade and no cancel. The `order_locations` entry survived that removal: `cancel_order` returned `Ok(None)` while re-adding the id reported `DuplicateOrderId`, so the id was permanently unusable and the tracked status stayed `Open` forever. For a two-tranche order the same update produced a `0 visible / N hidden` ghost.

## Changes

- **A zero `new_quantity` on `UpdateQuantity` cancels the order** rather than resizing it to nothing: the arm routes to `cancel_order_with_reason(UserRequested)`, the removal `OrderBook::cancel_order` performs, so the level-change event, the `Cancelled { UserRequested }` transition, the per-account risk release, the `order_locations` / `user_orders` untrack and the empty-level removal all happen in the lockstep the rest of the crate expects. Ungated: `update_order` already holds the submit gate.
- **The whole order goes, hidden depth included.** `new_quantity` is the visible tranche for an iceberg or reserve, so a zero there is not a zero total; the contract is that a zero requested quantity is a removal of the entire order, which is what the previous behaviour failed to deliver when it left the hidden tranche resting behind an empty display.
- **The removal runs before the projected validator and the modify-aware risk check**, so neither a configured `min_order_size` nor a risk limit vetoes it. The kill switch still refuses it, because it is submitted as a modify.
- The queue-priority contract on `update_order` documents the zero case alongside the existing decrease and increase rules (#203), the `# Errors` paragraphs are scoped to a **nonzero** `UpdateQuantity`, and `doc/USER_GUIDE.md` carries the same exception in its two-tranche and modification sections.

## Technical decisions

- Cancel, not reject. Zero is a removal and a size floor has no business vetoing it; the old path did not reject either, it rested an unfillable order.
- The semantic belongs to `UpdateQuantity` alone. A zero quantity on `Replace` or `UpdatePriceAndQuantity` re-adds the order through validate-first, so after #232 an iceberg or an auto-replenishing reserve rests with a zero visible tranche and live hidden depth, while a reserve with `auto_replenish` off is rejected with `ZeroVisibleTranche` and keeps resting. The docs say so rather than claiming an equivalence that does not hold.
- `Ok(None)` keeps meaning only that the requested order is absent.
- The wider total-versus-visible split between `with_reduced_quantity` and pricelevel is older than this PR and is not widened here.

## Testing

`tests/unit/update_quantity_zero_tests.rs`: the zeroed order leaves the book with `Cancelled { UserRequested }` and takes its now-empty level with it; the id is reusable (`cancel_order` reports it absent, a re-add succeeds); a post-only at the vacated price is admitted; a shared level keeps its other makers; zeroing an absent order stays `Ok(None)`; the update is refused under an engaged kill switch; an iceberg and a reserve with hidden depth are each cancelled whole; and a configured `min_order_size` does not veto the zero update.

Nineteen tests in total. Beyond the nine above: the #232 strandable-maker count drops by exactly one per cancelled non-auto reserve (two makers resting, so a double decrement would show); a configured risk limit does not veto the removal while the same book still rejects a resize that breaches it; the cancelled order's per-account contribution is released, proved by a second order the limit refuses before the cancel and accepts after it; the level-change event fires once on the zero path, on a level that survives and on one the removal empties; and the documented contrast with the other variants is asserted rather than described, `Replace` and `UpdatePriceAndQuantity` leaving an iceberg resting at zero visible with live, still fillable hidden depth, refreshing an auto-replenishing reserve out of hidden, rejecting a non-auto reserve with `ZeroVisibleTranche`, and removing a `Standard` maker with `Filled { filled_quantity: 0 }`.

`tests/unit/reserve_residual_policy_tests.rs` is updated for the interaction with #232: the zero `UpdateQuantity` case moves out of the `ZeroVisibleTranche` rejection loop, because the removal runs before the projected-shape validator ever sees it.

Full suite on this head: 871 lib + 598 integration + 26 fee + 6 metrics + 1 alloc-budget + 49 doc, 0 failed; `make test` 738 + 26 + 560 + 46, 0 failed. `cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

Rebased onto `main` after #232 and #222. Because #232 landed first, its release notes described a world where every quantity-carrying modify projection is validated: the changelog, `src/lib.rs` and `README.md` said `UpdateQuantity` with a zero quantity returns `ZeroVisibleTranche` with the original left resting. Those passages are scoped here to `Replace` and `UpdatePriceAndQuantity`, the variants the rule still reaches, and `lib.rs` gains the #223 What's New entry (README regenerated). Three further statements are corrected against the code: `is_zero_visible_ghost` matches a non-auto-replenishing reserve alone rather than every two-tranche kind, the restore prepare phase rejects that shape atomically so a legacy snapshot cannot hold one, and a `Standard` maker replaced with a zero quantity is cancelled and reported `Filled { filled_quantity: 0 }`.

`CHANGELOG.md` carries an `[Unreleased]` entry: this changes public contract behaviour on `update_order`. No public API change, no snapshot or journal format change. Replay note: a journal holding a zero `UpdateQuantity` replays to a removed order where it previously replayed to a resting one, so the replayed book legitimately differs and `ReplayEngine::verify` may return `Ok(false)`.


src/orderbook/modifications.rs
```diff
@@ -385,6 +390,23 @@ where
...
+                // nothing to fill, made `will_cross_market` reject a post-only
+                // at that price, and was eventually dropped by a sweep with no
+                // trade and no cancel event, leaking its `order_locations`
+                // entry (`cancel_order` then returned `Ok(None)` while a
+                // re-add of the id reported `DuplicateOrderId`). Cancel it
+                // instead — the same removal `OrderUpdate::Cancel` performs,
+                // and the one `Replace` / `UpdatePriceAndQuantity` already
+                // reach with a zero quantity. Ungated: `update_order` holds
+                // the shared submit gate (#209).
+                if new_quantity.as_u64() == 0 {
```
joaquinbejar, line 651: 🟠 **Warning**: The description says a configured `min_order_size` still rejects a zero update through the projected validator, but this return sits above `validate_order_shape(&projected)`, so it does not. I ran both branches with `set_min_order_size(5)` and a 10 lot ask. On main the update comes back `OrderSizeOutOfRange { quantity: 0, min: Some(5) }` and the maker stays. Here it comes back `Ok(Some)` and the maker is `Cancelled { UserRequested }`.

Cancelling is the behaviour I would keep, zero is a removal and a size floor has no business vetoing it, but then the description and the queue priority contract above should say so and a test should pin it. If you actually want the reject, move the branch below the validator.

src/orderbook/modifications.rs
```diff
@@ -243,6 +243,11 @@ where
     ///   pricelevel < 0.9 restore a demoted order at its old
     ///   `(timestamp, seq)` position — re-snapshot to pin the corrected
     ///   order.
+    /// - [`OrderUpdate::UpdateQuantity`] with a **zero** total quantity
+    ///   cancels the order: it is removed from the book, tracked as
+    ///   `Cancelled { UserRequested }`, and its id becomes reusable.
+    ///   A zero-quantity maker can never fill, so resting one only
+    ///   published a price level with no depth.
```
joaquinbejar: 🟠 **Warning**: This reads as a total but the check at line 406 is on the raw `new_quantity`, and upstream documents that field as the visible tranche only, with the hidden tranche left alone. For a two tranche order those disagree. I ran an iceberg of 10 visible and 50 hidden through `UpdateQuantity { new_quantity: 0 }` on both branches. On main it survives as a 0 visible 50 hidden ghost pinning `best_ask` with status `Open`, which is its own bug. Here the whole order is cancelled and the 50 hidden goes with it.

Cancelling is the policy I would keep, but the contract should say plainly that a zero `new_quantity` cancels the entire order, hidden depth included, and tests should pin it for iceberg and reserve. Zero visible quantity does not imply zero total quantity when hidden liquidity remains. The wider total versus visible split between `with_reduced_quantity` and pricelevel is older than this PR and is the same thing #221 reports on the price and quantity path, so I would not widen this PR to fix that.

```suggestion
    /// - [`OrderUpdate::UpdateQuantity`] with a **zero** `new_quantity`
    ///   cancels the entire order, including hidden quantity for iceberg
    ///   and reserve orders. It is removed from the book, tracked as
    ///   `Cancelled { UserRequested }`, and its id becomes reusable.
    ///   This applies even when hidden liquidity remains.
```

tests/unit/update_quantity_zero_tests.rs
```diff
@@ -0,0 +1,166 @@
...
+            order_id: Id::from_u64(MAKER),
+            new_quantity: Quantity::new(0),
+        })
+        .expect("update to zero succeeds")
+    }
+
+    /// The order leaves the book with a terminal cancel, and the level it
+    /// was alone on is removed with it.
+    #[test]
+    fn update_to_zero_cancels_the_order_and_removes_the_level() {
```
joaquinbejar, line 68: 🟠 **Warning**: Two cases are missing here, the ones where this branch actually changes the answer versus a plain resize. An iceberg or reserve order zeroed with hidden depth still resting, and a book with `set_min_order_size` where the zero update used to be rejected. Both pass today only because nothing asserts them.

joaquinbejar (changes requested):
The fix itself is right for a plain limit order and the plumbing is clean, the inner cancel is ungated so no gate reentry, the kill switch still runs first, state and risk release happen in lockstep and the empty level goes with it. Two things need sorting before merge. The description claims `min_order_size` still rejects a zero update and the code does the opposite, and two tranche orders are not covered while the branch cancels their hidden depth, both reproduced on the branch and on main. Please also open an [Unreleased] section in CHANGELOG.md for this, it changes public contract behaviour on `update_order`. REQUEST_CHANGES.

src/orderbook/modifications.rs
joaquinbejar: 🟠 **Warning**: One clarification to the documentation request. The zero branch bypasses the entire projected shape validator and modify-aware risk check, not just `min_order_size`. The `# Errors` paragraphs at lines 263 to 277 should therefore apply explicitly to nonzero quantity updates. A zero update cancels the whole order, while the existing kill-switch check still applies.

Please also qualify the new inline comment claiming the same behavior as zero-quantity `Replace` and `UpdatePriceAndQuantity`. Those paths still validate before cancelling, so the equivalence is not unconditional. This can remain a documentation and regression-test correction without expanding the PR into the older total-versus-visible issue.

Swofty-Developments:
I've kept the cancel policy and made the contract, errors section, and inline comment accurately document as such, including hidden depth and the validator bypass. Iceberg, reserve, and min_order_size are pinned.

joaquinbejar (commented):
Re-reviewed the latest head: the implementation matches the previously approved policy (zero UpdateQuantity cancels the whole order, including hidden depth, bypasses projected validation, and remains kill-switch gated). All 9 new tests pass; no implementation blocker found. Remaining corrections are documentation: the inline points below, plus doc/USER_GUIDE.md's two-tranche and modification sections, which still state unconditional hidden preservation and projected validation. Add the zero-UpdateQuantity exception there. Rewrite the PR description: its min-size and no-CHANGELOG statements describe the old version.

src/orderbook/modifications.rs
```diff
@@ -409,6 +426,30 @@ where
...
+                // zero-quantity maker in its queue (`new_total <= live_total`
+                // ⇒ keep in place), so applying the update rested a maker at
+                // zero depth: it held `best_bid` / `best_ask` on a level with
+                // nothing to fill, made `will_cross_market` reject a post-only
+                // at that price, and was eventually dropped by a sweep with no
+                // trade and no cancel event, leaking its `order_locations`
+                // entry (`cancel_order` then returned `Ok(None)` while a
+                // re-add of the id reported `DuplicateOrderId`). Cancel it
+                // instead — the same removal `OrderUpdate::Cancel` performs.
+                // `Replace` / `UpdatePriceAndQuantity` with a zero quantity
```
joaquinbejar: 🟡 Documentation — Zero requested quantity is not necessarily zero total quantity. Please change the opening 'zero total' at line 429 to 'zero requested quantity', scope the phantom/unfillable-maker rationale to zero-total orders, and remove or qualify this comparison with Replace/UpdatePriceAndQuantity. Independent probes on this head show that an iceberg updated with UpdatePriceAndQuantity(..., 0) remains live as 0/50 and subsequently fills those 50 hidden units; a reserve replaced with quantity 0 similarly remains 0/40 and fills its hidden depth. The approved whole-order cancellation policy for UpdateQuantity is fine, but these other variants do not provide the same removal semantics for two-tranche orders.

src/orderbook/modifications.rs
```diff
@@ -271,21 +280,29 @@ where
...
+    /// `UpdateQuantity` like they already were on the #98 modify paths:
+    /// an expired-but-unevicted GTD / DAY maker (`InvalidOperation`,
+    /// expiry is evaluated against the book clock) and a resting
+    /// post-only maker whose price meanwhile crosses the market
+    /// (`PriceCrossing`).
+    ///
+    /// A **zero** `UpdateQuantity` is a removal and runs none of that:
+    /// it bypasses the projected shape validator and the modify-aware
+    /// risk check entirely, so neither a configured `min_order_size` nor
+    /// a risk limit vetoes it, and takes the same cancel path as
```
joaquinbejar: 🟡 Documentation — The shared cancellation path here is `OrderBook::cancel_order` / `cancel_order_with_reason`, not the `OrderUpdate::Cancel` enum arm. That arm currently has a separate implementation and does not perform the same risk/tracker cleanup. Please correct this reference and the corresponding inline/CHANGELOG wording. The pre-existing enum-cancel discrepancy can remain outside this PR.

joaquinbejar:
Heads-up for the rebase: `OrderBook::get_bids()` / `get_asks()` are removed on `main` for 0.13.0 (https://github.com/joaquinbejar/OrderBook-rs/pull/232, issue #228), because the `Arc<PriceLevel>` handles they returned allowed mutating a level outside the submit gate. The two assertions in `tests/unit/update_quantity_zero_tests.rs` that read `book.get_asks().is_empty()` can become `book.best_ask().is_none()` (or `book.create_snapshot(usize::MAX).asks.is_empty()` if you want the whole side).

One more interaction from the same PR (#230 side): `validate_order_shape` now rejects any two-tranche projection with `visible == 0 && hidden > 0` with the new typed `OrderBookError::ZeroVisibleTranche`, on `add_order` and on the `UpdateQuantity` / `UpdatePriceAndQuantity` / `Replace` projections. Your zero-is-cancel branch on `UpdateQuantity` runs before validation, so it is unaffected, but the sentences saying that a zero quantity on `Replace` / `UpdatePriceAndQuantity` "zeroes the visible tranche while the hidden depth stays live" (CHANGELOG, USER_GUIDE and the `update_order` docs) will need to say those variants are rejected with `ZeroVisibleTranche` and leave the original resting. The same PR adds `tests/unit/reserve_residual_policy_tests.rs`, whose `test_update_order_two_tranche_zero_quantity_rejects_and_preserves_original` asserts that arm for `UpdateQuantity` too; once your zero-is-cancel branch lands, that `UpdateQuantity` case should expect the cancel instead (the `UpdatePriceAndQuantity` and `Replace` cases stay as they are).

joaquinbejar (commented):
Re-reviewed a0af863. All 9 zero-UpdateQuantity regression tests pass. The findings from my previous review are resolved in the repository: the branch checks the requested quantity, documents whole-order cancellation including hidden depth, correctly names the cancel_order helper, scopes the zero-total phantom rationale, and documents the validator bypass and guide exception. No remaining implementation or repository-documentation finding from this review.

The PR description still needs rewriting before merge: it says a size floor rejects zero, calls this the OrderUpdate::Cancel path, equates zero Replace/UpdatePriceAndQuantity with cancellation, and says no CHANGELOG was added. Those statements now contradict the tested contract and diff. Describe zero UpdateQuantity as whole-order cancellation via cancel_order_with_reason, bypassing projected shape/risk validation while still respecting the kill switch; the other modification variants retain their own validate-first behavior.

Coordination with unmerged #232: migrate the get_asks() test assertions when its accessors disappear, and preserve this early zero-cancellation branch despite #232's new projected-shape validation. That is integration work, not a failure against this PR's current base.

CHANGELOG.md
```diff
@@ -59,6 +59,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `ORDERBOOK_SNAPSHOT_FORMAT_VERSION` is unchanged. Single-thread and
   contention measurements are reported in the pull request.
 
+- **`OrderUpdate::UpdateQuantity` with a zero quantity cancels the order
+  (#223).** A zero `new_quantity` was accepted and applied as a resize:
```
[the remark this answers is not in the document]
joaquinbejar: The changelog entry is correct, I asked for it in review because this changes public contract behaviour on update_order. The description said otherwise and has been rewritten to match the branch.

src/orderbook/modifications.rs
```diff
@@ -409,6 +426,30 @@ where
                 order_id,
                 new_quantity,
             } => {
+                // A zero total is a removal, not a resize. pricelevel keeps a
```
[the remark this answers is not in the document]
joaquinbejar: Fixed on baef80a, the comment now says zero requested quantity and the phantom maker rationale is scoped to the one tranche case where zero requested is also zero total.

src/orderbook/modifications.rs
joaquinbejar: Fixed on baef80a and 6640016. The wording is zero requested quantity now, the unfillable maker rationale is scoped to zero total orders, and the Replace and UpdatePriceAndQuantity comparison is gone. In its place the docs state what those variants actually do after #232, an iceberg or auto replenishing reserve rests with a zero visible tranche and live hidden depth while a non auto reserve is rejected with ZeroVisibleTranche.

src/orderbook/modifications.rs
joaquinbejar: Fixed on 1e78fc7, the docs and the inline comment now name cancel_order_with_reason and OrderBook::cancel_order rather than the OrderUpdate::Cancel arm, and note that the enum arm is a separate implementation. The pre existing discrepancy there stays outside this PR.

joaquinbejar (approved):
All review items are addressed on d96ba08. The zero UpdateQuantity removal is correct and the docs no longer contradict #232, which landed first and described a world where every quantity carrying projection is validated. Ten tests were added for the paths the removal bypasses, the strandable maker count, risk release, the level change event and the contrast with Replace and UpdatePriceAndQuantity. Three stale claims inherited from #232 were corrected against the code. CI is green.

PR #224, opened 2026-09-09
Re-apply journaled submits that traded before returning an error

Swofty-Developments:
## Summary

`add_order` emits real fills and *then* returns `Err`, for an IOC's unfillable remainder, for a taker that STP cancels after non-self fills, and for a residual admission that fails once the sweep is irreversible. A sequencer records those as rejections, and `ReplayEngine::apply_event` skipped every rejected event, so replay silently rebuilt liquidity the live book had consumed: journal a resting ask of 10 and an IOC buy of 15, and the replayed book still carries the ask.

## Changes

- **`SequencerResult::RejectedWithCode { reason, code, may_have_mutated, stp_mode }`** (new, appended variant on the `#[non_exhaustive]` enum, so existing journals decode unchanged) and `impl From<&OrderBookError> for SequencerResult` to build it from the typed error. The string-only `Rejected` carries nothing replay can decide by, and no producer in the crate constructs it.
- **Replay decides by the recorded code, not by the success/failure classification.** A submit (`AddOrder`, `MarketOrder`, `MarketOrderByAmount`) journaled as `RejectedWithCode` is re-executed when its code is one replay can reproduce from the book state and `ReplayBookConfig`: the two post-fill codes (`InsufficientLiquidity`, `SelfTradePrevention`) and the pure admission rejections (tick, lot, size band, duplicate id, missing user, post-only crossing), which re-derive the same no-op and double as a check that the config matches the source book. The re-execution must fail under the same `RejectReason`, or replay aborts with the new **`ReplayError::OutcomeMismatch { sequence_num, recorded, actual }`**.
- **A rejection that may already have mutated the book is always re-executed, whatever its code.** `add_order` returns `OrderBookError::PriceLevelError` when the post-trade residual admission fails under concurrent flow: the engine logs at ERROR, removes the level the attempt created empty, and returns with the fills irreversible. That maps to `RejectReason::Other`, which the code-driven rule skips, so the entry replayed as a no-op and rebuilt the consumed liquidity. Rejections now carry `may_have_mutated`, derived from the typed error by an exhaustive match, and it overrides the skip. The clock-dependent expired-at-admission `InvalidOperation` shares the same code, is not a post-mutation failure, stays unflagged and is still skipped.
- **The STP mode is checked before dispatch.** Matching reject codes do not imply the same mutation: a taker reaching a same-user maker returns `SelfTradePrevented` under `CancelTaker` and `CancelBoth` alike, while `CancelBoth` also cancels that maker, so a replay configured with the wrong mode accepted the verdict and built a different book. `SelfTradePrevented` carries the mode that decided it, the journal records it, and replay compares it against the book's own mode, failing with the new **`ReplayError::StpModeMismatch`**. Coverage is a lower bound and documented as one: only a journal that contains an STP rejection can be checked this way.
- **Codes whose trigger lives outside the config are skipped:** the kill switch, the risk limits (`RiskConfig` is not part of `ReplayBookConfig`) and `Other` without the mutation flag. Those rejections never touch the book, so the skip reproduces the live outcome exactly. The skip list says so explicitly rather than leaving it to be inferred: a replay configured without them does not surface as `OutcomeMismatch`.
- The string-only `Rejected` keeps the historical skip. A rejected non-submit keeps the skip (the modify paths validate first, cancels are no-ops on a missing order). A journaled success whose re-execution fails still aborts with `ReplayError::OrderBookError`.
- **`last_applied_seq`, the applied count and the progress callback follow what replay dispatched to the book**, so a re-executed rejection advances them. Applied means dispatched, not that the book changed, and the count can exceed the number of events that changed anything; `replay_from` and both progress entry points state that.
- `[Unreleased]` entries in `CHANGELOG.md`: the fix under `### Fixed`, the API additions under `### Added`, the breaking variants under the existing 0.13.0 `### Changed` subsection.

## Technical decisions

- Nothing is replayed from the journaled result: replay re-executes matching deterministically, so the fills behind a tolerated error are reproduced by the re-execution itself. The recorded code is read only to decide whether to dispatch and to check the verdict.
- Journals are expected to record the outcome the command API returned. A success recorded for a command that returned `Err` is not a supported convention and aborts as a success/failure disagreement.
- The post-mutation flag is derived from the typed error rather than from the code, because the code is a wire value shared by unrelated conditions. It is an unreleased field on an unreleased variant and governs the event stream rather than the snapshot package, so `ORDERBOOK_SNAPSHOT_FORMAT_VERSION` is unchanged.
- **Stated limitations.** A journal written with the string-only `Rejected` keeps the pre-existing gap for traded-then-rejected submits; producers close it by recording `RejectedWithCode`. Only the code is compared, never the error's details or the fills behind it, and a discrepancy confined to them can go undetected, since different fills can exhaust the same levels and leave identical snapshots; `snapshots_match` is the check that exposes it. `MarketOrder` / `MarketOrderByAmount` carry no user id, so a market order journaled after STP effects under a user cannot be represented; it re-executes through the STP-less path and a rejection recorded for it aborts with `OutcomeMismatch` rather than diverging silently. Carrying the user on those commands is a follow-up.
- `OutcomeMismatch` and `StpModeMismatch` are new variants on the exhaustive `ReplayError`, a pre-1.0 breaking change. #228 already moved the crate to 0.13.0 as the release boundary, so this PR folds them into it: the bullet joins the breaking subsection of `CHANGELOG.md`, the migration note joins the 0.13.0 What's New list in `src/lib.rs`, and `README.md` is regenerated from it. Producers opt in with `SequencerResult::from(&error)`; string-only rejections keep their skip gap; exhaustive `ReplayError` matches need new arms.

## Testing

`tests/unit/replay_submit_error_tests.rs`, 19 tests, each journal built the way a sequencer does (run the command against a live book, record the outcome it returned): the IOC remainder and the STP-cancelled taker replay their fills and advance the applied sequence; the userless market commands replay their no-fill rejections as the same no-op; a kill-switch rejection is skipped with matching books and no progress callback; a tick rejection replays as the same no-op with the tick in the config and aborts with `OutcomeMismatch` without it; a rejection that fails under a different code aborts with both verdicts in the message; the string-only `Rejected` keeps the skip; a success journaled for a failed submit aborts with `OrderBookError`; a duplicate id journaled as a success still aborts; the progress callback sees the re-executed rejection.

Added this round: a post-mutation rejection recorded under an `Other` code is re-executed and aborts with `OutcomeMismatch { recorded: Other(0), actual: None }`; a pre-mutation `Other` rejection is still skipped with matching books and `last_applied == 0`; `may_have_mutated` is set for exactly the four post-mutation errors and unset for the other seven; a journal whose recorded STP mode is `CancelTaker` replayed under `CancelBoth` fails with `StpModeMismatch`; a journal with no recorded mode leaves the divergence to `snapshots_match`, which reports `false`; and `RejectedWithCode` round-trips through JSON and Bincode with all four fields intact.

Full suite on this head: 871 lib + 617 integration + 26 fee + 6 metrics + 1 alloc-budget + 49 doc, 0 failed. `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `make readme` clean.

Rebased onto `main` after #232, #222 and #223.


src/orderbook/sequencer/replay.rs
```diff
@@ -733,6 +752,40 @@ where
...
+    ///
+    /// Every other error means replay diverged from the recorded execution
+    /// and is surfaced as [`ReplayError::OrderBookError`].
+    fn reconcile_submit(
+        event: &SequencerEvent<T>,
+        rejected_live: bool,
+        outcome: Result<(), OrderBookError>,
+    ) -> Result<(), ReplayError> {
+        match outcome {
+            Ok(()) => Ok(()),
```
joaquinbejar: 🔴 **Critical**: A journaled `Rejected` submit whose re-execution returns `Ok` is a divergence. Live under `engage_kill_switch`, an ask is rejected without mutation, but replay here rests it and returns `Ok` with `snapshots_match` false. On main the event was skipped and the books matched. This introduces a silent divergence for previously harmless rejections.

My earlier proposed guard is necessary but insufficient. Seed an ask of 10, engage the kill switch, then journal an IOC buy of 15 as `Rejected` with `KillSwitchActive`. Replay consumes the ask and returns `InsufficientLiquidity { requested: 15, available: 10 }`, which `Err(_) if rejected_live` also accepts. Both executions failed, but for different reasons and with different effects. Check the recorded rejection against the replay error, preferably using a stable machine-readable category and relevant details, and abort on disagreement. Use a contextual divergence error carrying the sequence and recorded/actual outcomes. Reusing `SnapshotMismatch` would misdescribe this check; the earlier one-line suggestion has been withdrawn rather than presented as a complete fix.

src/orderbook/sequencer/replay.rs
```diff
@@ -733,6 +752,40 @@ where
...
+            Err(_) if rejected_live => Ok(()),
+            Err(
+                OrderBookError::InsufficientLiquidity { .. }
+                | OrderBookError::InsufficientLiquidityNotional { .. }
+                | OrderBookError::SelfTradePrevented { .. },
+            ) => Ok(()),
```
joaquinbejar: 🟠 **Warning**: Tolerating these three errors under every non-rejected classification removes divergence detection. A journal that says `OrderAdded` for an IOC buy of 10 against an ask of 5 replays here as `Ok`, while main aborts at that sequence. The new success-classification test discards the actual `Err` and manually records `OrderAdded`, contradicting its own `sequence` helper. That fixture does not justify this blanket allowlist.

One correction to my earlier wording. An installed `TradeListener` can receive a fee-bearing `TradeResult` before `add_order` returns `Err`, so a custom producer could record `TradeExecuted` from that callback. If that journal convention is supported, define it explicitly and validate the recorded trade outcome against replay, with a representative fixture. It does not justify accepting an arbitrary `OrderAdded` entry or ignoring which result was actually recorded. For journals recording the command API outcome, require matching success/failure and verify rejected outcomes too.

src/orderbook/sequencer/replay.rs
```diff
@@ -663,21 +663,39 @@ where
...
+    /// mutated book. A sequencer records those as `Rejected`, so skipping
+    /// them resurrected liquidity the live book had consumed. Replay
+    /// re-executes matching deterministically, so those fills are
+    /// reproduced by the re-execution itself — see
+    /// [`Self::reconcile_submit`] for which of its errors are tolerated.
+    ///
+    /// Every other command is failure-atomic (the modify paths validate
+    /// before touching the book, cancels are no-ops on a missing order), so
+    /// a `Rejected` result still skips it.
     fn apply_event(book: &OrderBook<T>, event: &SequencerEvent<T>) -> Result<(), ReplayError> {
```
joaquinbejar: 🟠 **Warning**: `replay_into` still computes `applied` from the journaled classification at line 622. Seed an ask of 10 at sequence 0 and journal an IOC buy of 15 as `Rejected` at sequence 1. Replay consumes the ask but returns `last_applied_seq` 0, and the progress callback omits the command that changed the reconstructed book. The returned applied sequence is now behind the work replay actually performed.

Have `apply_event` return `Result<bool, ReplayError>`, with true meaning the command was dispatched and its outcome reconciled, and false meaning a rejected non-submit was skipped. Use that result for the last applied sequence, event count and progress callback, including pure rejections and successful no-ops. Describe these as applied commands, not detected mutations, and test the final sequence and callback for a terminal rejected submit. This fixes the applied-command contract; it is not by itself a snapshot/resume API or a last-read journal cursor.

tests/unit/replay_submit_error_tests.rs
```diff
@@ -0,0 +1,310 @@
...
+        "the non-self maker was consumed live"
+    );
+    let config = ReplayBookConfig::new(None, STPMode::CancelTaker, None, None, None, None);
+    assert_replays_to(&live, &journal, &config);
+}
+
+/// A market order that only cancels same-user makers under `CancelMaker`
+/// mutates the book and still reports no liquidity.
+#[test]
+fn cancel_maker_market_rejection_replays_its_cancels() {
```
joaquinbejar: 🟠 **Warning**: This passes for the wrong reason. Replay runs `MarketOrder` through `submit_market_order` with no user, so STP never engages and the same user maker is filled by a trade instead of cancelled. On the replayed book `last_trade_price()` is `Some(100)` while live it is `None`, and `snapshots_match` cannot tell because it only compares levels. The gap is older than this PR, `MarketOrder` and `MarketOrderByAmount` carry no `user_id`, but the test claims coverage of maker cancels it does not have. With the abort from the `reconcile_submit` comment this case would fail loudly, which is the honest outcome until the command carries the user.

joaquinbejar (changes requested):
Re-executing rejected submits fixes the IOC and STP fill-replay gap, but reconciliation still silently accepts divergent executions. A kill-switch rejection can replay either successfully or as a different error after consuming liquidity, and the blanket error allowlist also accepts a false OrderAdded classification. Applied sequence and progress omit rejected submits that ran, while the CancelMaker market fixture trades instead of cancelling because its command lacks the live user identity. Tighten outcome reconciliation and these regressions, update the public replay contract, and add an [Unreleased] entry to CHANGELOG.md. REQUEST_CHANGES.

src/orderbook/sequencer/replay.rs
```diff
@@ -233,6 +234,34 @@ pub enum ReplayError {
...
+    /// same way, so a disagreement means the reconstructed book has
+    /// diverged from the live one — typically a [`ReplayBookConfig`] that
+    /// does not match the source book, a journal that does not start at
+    /// the book's origin, or state the journal cannot carry (a market
+    /// order's user identity).
+    #[error(
+        "replay diverged at sequence {sequence_num}: journal recorded rejection `{recorded}`, replay {}",
+        describe_outcome(.actual)
+    )]
+    OutcomeMismatch {
```
joaquinbejar, line 255: 🔴 P1 — This breaks the public exhaustive ReplayError API while Cargo.toml remains 0.12.1. The semver job fails specifically with `enum_variant_added` for `ReplayError::OutcomeMismatch`: https://github.com/joaquinbejar/OrderBook-rs/actions/runs/34690990214/job/103558100928 . A CHANGELOG note alone does not resolve downstream exhaustive-match breakage. Retain the contextual error but coordinate a breaking 0.x minor release (0.13.0), or choose a compatible public error representation. Do not suppress the check or treat this as a patch-only additive change.

src/orderbook/sequencer/replay.rs
```diff
@@ -731,7 +855,85 @@ where
...
+    ///   [`RejectReason`]; a success, or a failure under a different code,
+    ///   is [`ReplayError::OutcomeMismatch`]. The fills the live command
+    ///   made before failing are reproduced only if the verdicts agree, so
+    ///   a disagreement means the reconstructed book has diverged.
+    ///
+    /// Only the code is compared, not the error's details (`requested` /
+    /// `available` on `InsufficientLiquidity`, say): the journal carries
+    /// the code, and a divergence confined to the details is a divergence
+    /// in the book that the next dispatched command or a final
+    /// [`snapshots_match`] surfaces.
```
joaquinbejar: 🟡 Documentation — Matching RejectReason codes does not guarantee identical historical execution, and a later command or snapshots_match need not expose a discrepancy confined to error details. Different fills can exhaust the same levels and leave identical final snapshots. Please replace the detection guarantee here with an explicit limitation: details/fills are not compared and such discrepancies may remain undetected. The bounded code-comparison policy is already documented elsewhere; this paragraph should not promise stronger verification.

joaquinbejar (commented):
Re-reviewed the latest head: all 12 new replay tests pass. Independent probes also pass for a risk rejection without liquidity consumption, RejectedWithCode Bincode round-trip, and persisted IOC rejection replay after reopening FileJournal. The earlier kill-switch divergence, mismatched rejection-code acceptance, and applied-sequence accounting findings are resolved. The semver failure is a real remaining merge blocker (inline below). Please rewrite the PR description: it still describes unconditional replay of string-only Rejected, blanket error tolerance, and unchanged counters, none of which matches this head. Also add the required public-surface migration note in src/lib.rs and regenerate README: producers opt in with SequencerResult::from(&error), legacy string-only rejections retain their skip gap, older readers cannot decode the new variant, and exhaustive ReplayError matches need migration.

joaquinbejar:
Heads-up on the release boundary. https://github.com/joaquinbejar/OrderBook-rs/pull/232 (issue #228) also moves `Cargo.toml` to 0.13.0 and adds the 0.13.0 "What's New" section in `src/lib.rs` (README regenerated from it) plus a `### Changed (breaking, semver-minor under 0.x)` subsection under `[Unreleased]` in `CHANGELOG.md`. Once it lands you will hit conflicts on those four files when rebasing. Keep the single 0.13.0 bump, fold the `ReplayError::OutcomeMismatch` bullet into that same breaking subsection and into the existing 0.13.0 "What's New" list rather than adding a second section, and run `make readme` so `README.md` matches `lib.rs`. 0.13.0 is the boundary for both breaking changes, so nothing else about the version needs to change on your side.

joaquinbejar (commented):
Re-reviewed d3f674f. My previous semver and verification-limit findings are resolved: Cargo.toml now targets 0.13.0, the semver job is green, lib.rs/README carry the public migration note, and the reconciliation docs explicitly acknowledge that equal rejection codes and equal final snapshots do not prove identical historical execution. Executable replay/type code is unchanged from the previously reviewed 919daed implementation. All 12 replay regressions and three independent probes (risk-rejection skip, Bincode round-trip, and replay after reopening FileJournal) pass. No remaining material code finding from this review.

The PR description still describes the abandoned implementation. Please rewrite it around RejectedWithCode opt-in: legacy string-only Rejected stays skipped; reproducible coded submit rejections are re-executed and must produce the same code; kill-switch/risk/Other codes are skipped; a success/error disagreement aborts; applied sequence/count/progress follow dispatched commands. Remove the claims that counters remain classification-based, that arbitrary classifications tolerate post-fill errors, and that no CHANGELOG was added. Include the 0.13.0 compatibility boundary and the code-only verification limitation.

When integrating #232, retain both sets of release notes and test registrations under the same 0.13.0 boundary. Its added rejection mappings do not require changing this replay classifier.

CHANGELOG.md
```diff
@@ -96,6 +96,66 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   resulting snapshot differs from the one captured by the pre-fix run,
   or propagate the replay error when the re-executed update is rejected.
 
+- **Replay re-executes journaled submits that traded before returning an
```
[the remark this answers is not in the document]
joaquinbejar: The changelog entry is correct and stays, the description was stale and is being rewritten with the rest of this round.

src/orderbook/sequencer/replay.rs
joaquinbejar: Resolved. The branch is rebased on main, which is already 0.13.0 after #232, and the OutcomeMismatch variant is folded into the 0.13.0 release notes as a breaking change on the same boundary rather than announced as additive. The semver job runs against that base now.

src/orderbook/sequencer/replay.rs
joaquinbejar: Fixed on 29943a6 and b94b22b. The paragraph now says the reconciliation compares reject codes only, that a discrepancy confined to details or fills can go undetected because different fills can exhaust the same levels, and that snapshots_match is what exposes it.

src/orderbook/sequencer/replay.rs
```diff
@@ -731,7 +855,85 @@ where
...
+    /// [`ReplayError::OutcomeMismatch`] rather than being skipped silently.
+    #[must_use]
+    fn replays_rejection(code: RejectReason) -> bool {
+        !matches!(
+            code,
+            RejectReason::KillSwitchActive
+                | RejectReason::RiskMaxOpenOrders
+                | RejectReason::RiskMaxNotional
+                | RejectReason::RiskPriceBand
+                | RejectReason::Other(_)
```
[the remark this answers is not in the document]
joaquinbejar: Confirmed and fixed on bec9915. add_order returns PriceLevelError after its sweep emitted trades when the post trade residual admission fails, which maps to Other and was skipped, so the entry replayed as a no op and rebuilt consumed liquidity. Rejections now carry may_have_mutated, derived from the typed error by an exhaustive match, and it overrides the code driven skip; the clock dependent expired at admission case shares the same code, is not post mutation, and is still skipped.

src/orderbook/sequencer/replay.rs
```diff
@@ -731,7 +855,85 @@ where
...
+        recorded: Option<RejectReason>,
+        outcome: Result<(), OrderBookError>,
+    ) -> Result<(), ReplayError> {
+        match (recorded, outcome) {
+            (None, Ok(())) => Ok(()),
+            (None, Err(source)) => Err(ReplayError::OrderBookError {
+                sequence_num: event.sequence_num,
+                source,
+            }),
+            (Some(code), Err(actual)) if RejectReason::from(&actual) == code => Ok(()),
```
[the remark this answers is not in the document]
joaquinbejar: Fixed on bec9915. SelfTradePrevented carries the mode that decided it, so the journal records that mode and replay compares it against the book's own before dispatch, failing with the new ReplayError::StpModeMismatch. Coverage is a lower bound and documented as one, only a journal that contains an STP rejection can be checked this way, and the docs now say plainly that details and fills are never compared and that snapshots_match is what exposes such a discrepancy.

src/orderbook/sequencer/replay.rs
```diff
@@ -615,12 +700,14 @@ where
...
-            let applied = !matches!(event.result, SequencerResult::Rejected { .. });
-            Self::apply_event(book, event)?;
+            // correct even if the event is skipped. `last_applied_seq`,
+            // `count`, and `progress` track the events `apply_event`
+            // dispatched to the book — including a re-executed rejection,
+            // which may have traded before it failed — consistent with the
+            // "events applied" / "last applied sequence" contract on the
+            // public entry points. Applied means dispatched, not that the
+            // book changed.
+            let applied = Self::apply_event(book, event)?;
```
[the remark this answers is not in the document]
joaquinbejar: The behaviour is right and the wording was not. Applied means dispatched, so last_applied_seq and the applied count include a re executed rejection and can exceed the number of events that changed the book. That is now stated on replay_from and on both progress entry points, and the description is rewritten to match.

src/orderbook/sequencer/replay.rs
```diff
@@ -731,7 +855,85 @@ where
...
+    /// not list is re-executed, so a divergence surfaces as
+    /// [`ReplayError::OutcomeMismatch`] rather than being skipped silently.
+    #[must_use]
+    fn replays_rejection(code: RejectReason) -> bool {
+        !matches!(
+            code,
+            RejectReason::KillSwitchActive
+                | RejectReason::RiskMaxOpenOrders
+                | RejectReason::RiskMaxNotional
+                | RejectReason::RiskPriceBand
```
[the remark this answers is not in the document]
joaquinbejar: Reconciled in favour of the implementation. Kill switch and RiskConfig rejections are skipped rather than re executed, so a replay configured without them does not surface as OutcomeMismatch; the skip list documents that explicitly now and the description no longer promises otherwise.

joaquinbejar (approved):
All seven threads are addressed on 852380a. The Other skip no longer hides a rejection that already mutated the book, the recorded STP mode is checked before dispatch with StpModeMismatch, and the applied count and skip list now say what the code does rather than what the description promised. The docs state plainly that only reject codes are compared and that snapshots_match is what catches a discrepancy confined to details or fills. Rebased onto main after #232, #222 and #223; CI including semver is green.

Issue #226, opened 2026-09-12
Validate Reserve lot size per tranche like Iceberg (orderbook/modifications)


joaquinbejar:
## Overview

`OrderBook::validate_order_shape` (`src/orderbook/modifications.rs`, lot-size block) validates the two two-tranche order kinds differently. `OrderType::IcebergOrder` gets a dedicated arm that checks `visible_quantity % lot` and `hidden_quantity % lot` separately, while `OrderType::ReserveOrder` falls into the `_` arm together with the single-tranche kinds and is checked on `total_quantity() % lot` only. With `lot_size = 10`, a reserve of 15 visible / 5 hidden is admitted (total 20) while the identical iceberg is rejected with `InvalidLotSize`. Since #221 a reserve `UpdatePriceAndQuantity` actually applies the requested visible quantity, so the projected reserve order reaches this check with sizes it never reached before, which makes the asymmetry observable through the public modify path as well as through `add_order`.

The `_` catch-all on `OrderType` is also the pattern `rules/global_rules.md` ("Pattern matching") flags: a future variant would silently inherit the total-only rule.

## Context

- `src/orderbook/modifications.rs`, `validate_order_shape`, the `if let Some(lot) = self.lot_size` block: `IcebergOrder` arm vs `_` arm.
- Reserve-specific fields that may need a lot-size opinion: `replenish_amount` (`Option<NonZeroU64>`) and `replenish_threshold` (a refill that is not a lot multiple would surface non-lot visible sizes later).
- Related: #210 (two-tranche conservation, `QuantityOverflow` check at the top of the same function), #221 (visible-tranche semantics on `set_quantity`).

## Tasks

- [ ] Make the lot-size rule identical for `IcebergOrder` and `ReserveOrder` (per-tranche `% lot`, matching the iceberg arm), and decide whether `replenish_amount` / `replenish_threshold` must also be lot multiples.
- [ ] Replace the `_` arm with an exhaustive match over `OrderType` so a new variant must state its own lot rule.
- [ ] Unit tests in `src/orderbook/tests/modifications.rs` (or the validation tests under `tests/unit/validation_tests.rs`) covering: reserve with a non-lot visible tranche rejected, reserve with a non-lot hidden tranche rejected, lot-aligned reserve admitted, and the same three for iceberg, on both `add_order` and `UpdatePriceAndQuantity`.
[...]

Issue #225, opened 2026-09-12
Fix intermittent same-user trade under concurrent admission (orderbook/matching)


joaquinbejar:
## Overview

`tests/unit/atomic_postonly_fok_tests.rs::tests_post_only_stp_precedence::post_only_rejection_never_cancels_same_user_makers` fails intermittently at line 305 with `round N: zero trades under every interleaving` (`left: 1`). The round submits, from two threads released by a `Barrier`, a same-user GTC ask (sell 5 @ 100) and a same-user post-only buy (5 @ 100) on a book with `STPMode::CancelMaker`. No interleaving should ever produce a trade: a post-only never takes, and an ask that meets a resting same-user post-only must cancel it under `CancelMaker`. One trade is emitted anyway in a small fraction of runs, which means the STP guarantee is violated under concurrent same-user admission.

This is not a test-harness artefact of any open branch: on `main` (`e7331f0`) the full default-feature integration suite (`LOGLEVEL=WARN cargo test --test tests`, the exact `make test` / CI `run_tests` configuration) failed 1 of 40 runs with this assertion (`round 47`); on an unrelated branch it failed 3 of 40 (`round 154`, `32`, `185`) plus once inside `make pre-push`. Running the test alone (`cargo test --test tests post_only_rejection_never_cancels_same_user_makers`, 40 runs) never failed, so the race needs the CPU contention of the parallel suite to open. Every CI run of every PR currently carries a few-percent chance of going red on this test regardless of its diff.

## Context

Suspected window (hypothesis, not yet instrumented): in `OrderBook::match_order_with_user_outcome` (`src/orderbook/matching.rs`, STP pre-processing block) the per-level STP scan (`price_level.snapshot_by_seq_into(&mut stp_orders)` + `check_stp_at_level`) and the subsequent `price_level.match_order(...)` are two separate operations on the same level with no exclusive hold across them. A concurrent `add_order` from the same user that lands on that level between the two yields `STPAction::NoConflict` followed by a fill against the same-user maker. The #209 submit gate is a shared (read) acquisition for non-FOK adds, so both admissions run concurrently by design (`src/orderbook/modifications.rs`, `add_order_with_result`). The post-only side has the mirror shape: its crossability probe runs before the STP block and before the contra order has rested.

[...]

PR #232, opened 2026-09-12
Remove level handles, reconcile the reserve residual policy (#228, #230)

### LINKED ISSUE #228: Stop exposing mutable PriceLevel handles through get_bids / get_asks (orderbook/book)
## Overview

`OrderBook::get_bids` and `OrderBook::get_asks` (`src/orderbook/book.rs:3516` and `:3528`) return `Arc<DashMap<u128, Arc<PriceLevel>>>` holding the live `Arc<PriceLevel>` handles of the book's levels. `pricelevel::PriceLevel` exposes `add_order`, `update_order` and `match_order` publicly, so a caller can mutate a level directly through those handles. Every such mutation bypasses the book's invariants: the `submit_gate` (#209 / #225), `order_locations`, the risk layer, STP, the kill switch, the order-state tracker and both listeners. The concurrency guarantee documented for STP in #225 is therefore scoped to mutations performed through the `OrderBook` API; this issue tracks closing the hole in the API itself.

## Context

- `src/orderbook/book.rs`, `get_bids` / `get_asks`, and their callers in the crate (statistics, snapshots, iterators) and in `examples/` / `tests/`.
- `pricelevel::PriceLevel` public mutators: `add_order`, `update_order`, `match_order`.
- Related: #209 (submit gate), #225 (STP under concurrent admission, guarantee scoped to the `OrderBook` API), `rules/global_rules.md` "Type Safety" (public types with `pub` access that can violate invariants).

## Tasks

- [ ] Decide the replacement surface: read-only level views (for example `PriceLevelSnapshot` values, or a borrowed read-only wrapper over `PriceLevel` that exposes only queries), keeping the existing iterator / statistics / snapshot consumers working.
- [ ] Migrate internal callers, examples and tests off the mutable handles.
- [ ] Deprecate `get_bids` / `get_asks` first (with the read-only alternative documented) if a hard removal is too disruptive for 0.x consumers, then remove in the next minor.
- [ ] Document in `src/lib.rs` / `README.md` (via `make readme`) and `CHANGELOG.md` that all book mutation must go through `OrderBook`.

## Acceptance criteria

- [ ] No public `OrderBook` accessor hands out a handle through which a level can be mutated outside the submit gate.
- [ ] Existing depth / statistics / snapshot consumers keep their behaviour, with tests.
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `cargo test --all-features` clean
- [ ] `cargo build --release` clean
- [ ] Relevant docs updated (`src/lib.rs` docs, `README.md` regenerated via `make readme`, `CHANGELOG.md`)

## References

- #209, #225
- `rules/global_rules.md`, "Type Safety" and "Concurrency"

### LINKED ISSUE #230: Reconcile reduce_reserve_to_total with pricelevel's auto_replenish contract (orderbook/modifications)
## Overview

The crate applies two opposite replenishment policies to the same `ReserveOrder` depending only on whether it arrives crossing the book. `reduce_reserve_to_total` (`src/orderbook/modifications.rs`, the residual-resting helper behind `OrderQuantity::set_total_remaining`) refreshes an emptied visible tranche with `replenish_amount.min(hidden)` without consulting `auto_replenish`, so an aggressive reserve taker `{ visible 10, hidden 20, replenish_amount: Some(10), auto_replenish: false }` that fills 10 rests its residual as 10 visible / 10 hidden. The identical order resting first as a maker and then fully consumed is instead removed by pricelevel with its 20 hidden units intact and no replenishment, because pricelevel 0.9.1's `match_against` returns `(consumed, None, 0, remaining)` when `auto_replenish` is false, per its documented contract ("if `auto_replenish` is false and visible quantity is zero it will be removed from the book").

#226 validates the lot alignment of the transfer that `reduce_reserve_to_total` actually performs (an explicit `replenish_amount`, capped by hidden, regardless of `auto_replenish`), so admission is consistent with the current code; this issue is about making the two paths agree.

## Context

- `src/orderbook/modifications.rs`, `reduce_reserve_to_total` (called from `set_total_remaining`, used by `add_order_inner` when an aggressive residual rests), and the `# Lot size` section of `validate_order_shape` (#226), whose `(Some(amount), _)` arm exists because of this divergence.
- `pricelevel` 0.9.1, `src/orders/order_type.rs`: `ReserveOrder` docs (~line 172) and `match_against` (~lines 809-910).
- Sequencing note from the #226 review: making the helper honour `auto_replenish` naively would leave a residual resting at `visible 0 / hidden > 0`, which pricelevel then deletes on the next taker without a trade; the same change therefore needs a zero-visible guard (drop the residual, or rest nothing, when `auto_replenish` is off and the visible tranche is exhausted), and a decision on what happens to the hidden depth.
- Related: #210 (two-tranche conservation), #221 (visible-tranche semantics), #226 (lot-size validation of the transfer).

## Tasks

- [ ] Decide the single policy for a non-auto reserve whose visible tranche is exhausted while it still holds hidden depth, consistent with pricelevel's contract (removal, or an explicit re-display) and document it on `ReserveOrder` handling in `add_order` / `set_total_remaining`.
- [ ] Make `reduce_reserve_to_total` follow that policy, with the zero-visible guard, and keep quantity conservation (`executed + resting visible + resting hidden` accounted for, including hidden depth that is dropped).
- [ ] Revisit the `(Some(amount), false)` arm of the #226 lot-size rule once the helper no longer transfers without `auto_replenish`; keep the rule faithful to whatever the code then does.
- [ ] Tests: the aggressive-taker residual case for `auto_replenish = false` with and without an explicit amount, the maker depletion case, conservation and lot alignment on a lot-size book; update `tests/unit/reserve_lot_size_tests.rs`, whose residual test currently records the divergent 10 / 10 outcome.
- [ ] `CHANGELOG.md` entry describing the behaviour change for aggressive reserve residuals.

## Acceptance criteria

- [ ] An aggressive reserve taker and a resting reserve maker with the same fields follow the same replenishment policy when the visible tranche is exhausted.
- [ ] No residual ever rests at `visible 0 / hidden > 0`.
- [ ] Quantity conservation and lot alignment hold in the new tests.
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `cargo test --all-features` clean
- [ ] `cargo build --release` clean
- [ ] Relevant docs updated (`CHANGELOG.md`; `src/lib.rs` docs and `README.md` via `make readme` if the public wording changes)

## References

- #210, #221, #226
- `rules/global_rules.md`, "Matching / Book Discipline"

joaquinbejar:
## Summary

Two issues, one release boundary. **#230**: the residual-resting helper `reduce_reserve_to_total` applied a replenishment policy to an aggressive reserve taker that contradicted what pricelevel does to the identical resting maker on both counts (it refreshed from an explicit amount even with `auto_replenish` off, and refreshed nothing when the amount was `None` even with it on, resting a zero-visible order). The residual now follows `auto_replenish` exactly like the maker path. **#228**: `get_bids()` / `get_asks()` handed out live `Arc<PriceLevel>` handles through which a level could be mutated outside every book invariant; they are removed, which makes this the 0.13.0 breaking boundary.

### #230 — reserve residual policy

- `reduce_reserve_to_total` refreshes an emptied visible tranche only when `auto_replenish` is on, with `replenish_amount.unwrap_or(DEFAULT_RESERVE_REPLENISH_AMOUNT)` capped by hidden (pricelevel's rule); with it off the visible tranche stays empty.
- Scoped guard in `add_order_inner`: an exhausted non-auto reserve residual does not rest; the order ends as `Filled { filled_quantity: executed }`, mirroring the maker path, and the discarded hidden quantity is logged. Iceberg residuals and auto-replenishing reserves are untouched; no blanket zero-visible rule.
- The helper also mirrors pricelevel's below-threshold refresh: with automatic replenishment on and hidden depth left, a residual whose visible tranche drops below `safe_threshold` (the threshold, or 1 when it is 0) is topped up by the same capped transfer, so a residual and a maker consuming the same quantity display the same tranches.
- **Validate-first pre-check on the cancel-then-add modify variants (review blocker).** `UpdatePrice`, `UpdatePriceAndQuantity` and `Replace` re-add through the same code path, so a non-auto reserve re-priced across depth at least equal to its visible tranche would have been cancelled, executed and then discarded while `update_order` reported success. Each arm now dry-runs the projected order's crossable depth with the fill-or-kill feasibility helper and rejects the modify with the new typed `OrderBookError::ReserveResidualWouldBeDiscarded { order_id, visible_quantity, crossable_quantity, hidden_quantity }` before anything is cancelled. The rejection band is `crossable > 0 && crossable >= visible && crossable < total`: a projected full fill discards nothing and is allowed, zero crossable depth rewrites no tranche, and a re-price crossing less than the visible tranche executes and rests; auto-replenishing reserves and icebergs are unaffected. The error maps to the new additive wire code `RejectReason::ReserveResidualWouldBeDiscarded = 14` (older readers decode it as `Other(14)`) rather than the `Other(0)` sink. Since the review round the re-price of a strandable reserve holds the exclusive gate, so the dry run is exact for the only orders it applies to.
- **No zero-visible two-tranche orders.** `validate_order_shape` rejects an iceberg or reserve with `visible == 0 && hidden > 0` with the new `OrderBookError::ZeroVisibleTranche { order_id, hidden_quantity }` (wire code `RejectReason::InvalidQuantity`, 8), which covers admission, the `UpdateQuantity` / `UpdatePriceAndQuantity` / `Replace` projections that since #221 set the visible tranche, and the residual path's admission; it removes the discontinuity where a `{0, N, false}` reserve that never traded rested as a ghost displaying nothing and was removed with N stranded by the first taker. Nothing is discarded by this rule; the original order is untouched.
- **Defined return shape and observability of the discard.** The guard returns the ended order with both tranches at zero (`total_quantity() == 0`); `add_order` documents what each branch returns (fully matched: the submitted order; resting: the resting residual; discarded: the ended order). The discard is recorded under the `metrics` feature as two counters, `orderbook_reserve_discards_total` (orders) and `orderbook_reserve_hidden_discarded_total` (quantity units; neither derivable from the other), and traced at `info!` with the order id, the executed and the discarded quantities, on **both** paths: the taker residual guard and the maker removal in `matching.rs` where pricelevel strands a depleted non-auto reserve's hidden depth (trace field `path = maker | taker`). The `Filled` rustdoc states that `filled_quantity` counts executed quantity only and that a non-replenishing reserve's hidden remainder may have been discarded.
- **Cost model of the maker-side capture, measured.** pricelevel drops the stranded hidden quantity without reporting it, so `capture_strandable_makers` walks `PriceLevel::iter_orders()` (a `DashMap` iteration that read-locks every shard, 64 on the test machine) before a level is matched. The walk is armed by a monotonic per-book flag set the first time a non-auto reserve with hidden depth rests or is restored from a snapshot, and runs only on levels that hold hidden depth: books that never rest such a maker (plain, iceberg-only or auto-reserve books) pay one relaxed load per sweep and never take the buffer from the pool. Numbers in **Performance**; the flag is never cleared, since clearing it would need exact removal bookkeeping and a stale `true` only costs the walk while a stale `false` would lose reports. The honest long-term fix is upstream: pricelevel reporting the stranded quantity in its match result would remove the walk.
- New HDR bench `reserve_sweep_hdr` (two scenarios: strandable vs replenishing reserve makers), wired into `make bench-hdr` and documented in `BENCH.md` with the paired `main` vs branch table, so the capture's cost stays measurable.
- The #226 lot-size transfer check applies only with automatic replenishment on, because with it off nothing ever transfers in either path.
- Docs on `set_total_remaining`, the helper, `add_order` (two-tranche takers) and the validator; the unreleased #226 changelog bullet amended to the final rule. `set_total_remaining` and the user guide state the consequence of the default refresh `min(DEFAULT_RESERVE_REPLENISH_AMOUNT, hidden)`: the refreshed tranche can display more than the submitter originally showed (10 / 20 executing 10 rests 20 / 0).

### Review round (maintainer review + Copilot, 2026-09-12)

- **Coherent reporting boundary.** A book that holds strandable makers (`ReserveOrder { auto_replenish: false, hidden > 0 }`) runs every matching-capable submit and every cancel-then-add re-price under the **exclusive** side of the submit gate, and admitting a strandable maker is exclusive in every `STPMode`. The mode is decided from `strandable_makers_resting`, never from an order lookup: `acquire_coherent_submit_gate` takes the shared side, re-reads the count and, if it is no longer zero, drops the guard and takes the exclusive side afresh before the caller has read anything (a restart, not an upgrade). Invariant: no strandable maker can be admitted, cancelled or replaced while a sweep runs in a book that holds one, so a level's capture, its match and the report see one set of makers and ids; this also makes `check_modify_reserve_residual`'s dry run exact for the only orders it applies to. Post-only, `UpdateQuantity` and cancels stay shared. Cost: such books serialize like STP books since #225; books holding none pay one relaxed load per sweep.
- **Deterministic regressions** (`src/orderbook/tests/strandable_gate_concurrency.rs`, channel rendezvous, `#[cfg(test)]` level hook, `try_read` / `try_write` observing `WouldBlock`, timeouts only detect hangs): a strandable maker admitted while a sweep is parked blocks until it ends (pre-fix the sweep executed 3 instead of 2 and swallowed it); a cancel plus a Standard re-admission under a captured maker's id cannot land in the capture window (pre-fix the sweep filled the impostor and the cancel returned `Ok(Some(_))` inside the window); a re-price in such a book is exclusive and returns to shared once the count is zero.
- **Exact count** `strandable_makers_resting: AtomicUsize` replaces the monotonic flag: +1 at the two insertion points (`add_order_inner`, `commit_restored_levels` after a reset), −1 from the removed order's own body in the cancel funnel (user cancel, the three scoped mass cancels, expiry eviction, cancel-then-add modifies), the STP maker cancel and the fill drain, reset in `cancel_all_orders`. The drain decrement is exact because the exclusive sweep rules out a stale capture. Lookup: the captured makers are sorted once by `Id::as_bytes()` and binary-searched per filled id (`Id` has no `Ord`; the equal-key run is walked and compared by `Id` equality). Tests per removal path, cancel versus match decrementing exactly once, a reused id never decrementing, recount on direct, package and JSON restore.
- **Diagnostic**: `ReserveResidualWouldBeDiscarded` gains `discarded_quantity = total − crossable` next to the projected `hidden_quantity`; `Display` distinguishes them; asserted at depth 15 (15 of 20) and depth 10 (20 of 20).
- **`ZeroVisibleTranche` narrowed** to the non-auto reserve with visible 0 and hidden > 0, the only shape pricelevel removes without a trade (a zero-visible iceberg draws its whole hidden tranche into visible when hit; an auto reserve refreshes and re-queues). Zero-visible icebergs and auto reserves are admitted again and covered as makers (execute 20 / 10 when hit) and as partially filling takers (residuals `(0, 15)`, later matchable for 15, and `(10, 5)`). The #223 zero `UpdateQuantity` exception is documented as taking precedence for that variant.
- **Legacy ghost on restore**: `prepare_snapshot_levels` rejects the narrowed shape with `ZeroVisibleTranche` before any state changes, via the same `is_zero_visible_ghost` helper the validator uses; tested on direct, package and JSON restore with the destination's orders and configuration intact. Such snapshots must be repaired before restoration.
- **Docs**: the discard needs an exhausted visible tranche (10/20 filled for 5 rests 5/20); the explicit `replenish_amount` is the transfer added to the surviving visible tranche (amount 10, threshold 5, remainder 2 gives 12), the "pin the displayed size" advice is gone; the modify guarantee is scoped to a re-price of a strandable reserve under the exclusive guard; the band's non-crossing case is its own sentence.

### #228 — mutable level handles

- `OrderBook::get_bids` / `get_asks` removed (no callers in the crate, tests, examples or benches). Read-only alternatives, documented in `lib.rs`, the user guide and the changelog: `create_snapshot`, `levels_with_cumulative_depth`, `levels_until_depth`, `levels_in_range`, `find_level`, `order_count_at_price`, `get_orders_at_price`, `get_all_orders`, `total_depth_at_levels`, `best_bid` / `best_ask`.
- `Cargo.toml` bumped to 0.13.0; `lib.rs` gains the 0.13.0 "What's New" section (this removal plus #221, #225, #226, #230) and `README.md` is regenerated from it.
- The STP scope wording (`acquire_submit_gate`, `STPMode`, user guide, changelog) now says the public API hands out no level handles.
- The three `new` constructors in `orderbook::iterators` took `&SkipMap<u128, Arc<PriceLevel>>`, the book's internal representation; nothing public yields that reference after the removal, so they are `pub(crate)` inside this same breaking boundary (the iterator types stay public and are obtained from the `OrderBook` methods).
- `get_order_locations_arc` doc corrected (and `#[must_use]` added): it returns a fresh copy of the order-location map, not a handle to the live one.

## Technical decisions

- **One policy, both paths (#230).** `auto_replenish` decides the fate of a depleted visible tranche whether the order arrived crossing or rested first. With it on, the tranche is refilled with the explicit amount or pricelevel's default, capped by hidden. With it off, pricelevel removes a resting maker and discards its hidden depth; the aggressive residual now ends the same way instead of resting a refreshed or zero-visible order. `Filled { filled_quantity: executed }` is kept for compatibility with the maker path.
- **Accounting.** `submitted = executed + resting (visible + hidden) + discarded`, asserted in every residual test; discarded hidden quantity is never counted as executed and is logged when the guard fires.
- **Execution asymmetry preserved and documented.** An aggressive two-tranche order sweeps with its total quantity, so a 10 / 20 reserve submitted into 20 units of liquidity executes 20; the identical non-auto resting maker executes only its 10 visible units before pricelevel removes it. Reconciling the residual policy does not change this, and the tests assert both numbers.
- **Removal, not deprecation (#228).** Deprecation would have left the bypass available and failed the issue's acceptance criterion. Removing public functions is breaking, hence 0.13.0; that boundary is shared with the `ReplayError` variant addition under discussion in #224. PR #223's new tests call `book.get_asks().is_empty()`; the equivalent read-only assertion is `book.best_ask().is_none()` (or `create_snapshot(usize::MAX).asks.is_empty()`), noted on that PR.
- **Compatibility.** No snapshot or journal format change and no `ORDERBOOK_SNAPSHOT_FORMAT_VERSION` bump. A journal recorded before this change whose aggressive non-auto reserve residual rested (for example the old 10 / 10 outcome) now replays to that order ending after its fills, so later commands that touched the rested residual may produce different historical results. Snapshot restoration is unchanged.

## Performance

Hot-path changes under test: `capture_strandable_makers` and the exclusive-gate rule for books holding strandable makers (#230). Repo HDR harness, Apple M-series (12 performance + 4 efficiency cores) macOS host, release profile, `main` = `b821df2` (0.12.1) vs this branch (0.13.0), same `Cargo.lock`, pricelevel 0.9.1 on both, runs interleaved `main` / branch on an otherwise idle machine. p50 of the µs-range scenarios is bimodal run to run (a single-threaded run lands on either core type), so medians with the run range in brackets are reported, never single runs. Values in ns.

**Common path** (books that never rest a non-auto reserve with hidden depth: one relaxed load per sweep, no buffer, shared gate as before):

| bench | quantile | `main` | branch |
| --- | --- | --- | --- |
| `aggressive_walk_hdr` (18 paired runs) | p50 | 42 | 42 |
| | p99 | 3 793 [2 417..6 375] | 3 375 [2 501..6 791] |
| | p99.9 | 7 751 [5 087..13 751] | 7 043 [5 459..14 295] |
| `thin_book_sweep_hdr` (18 paired runs) | p50 | 83 | 83 |
| | p99 | 4 627 [4 001..6 211] | 4 459 [4 211..5 627] |
| | p99.9 | 6 791 [5 127..9 295] | 6 919 [5 419..8 375] |
| `add_only_hdr` (9 paired runs) | p50 | 1 125 [1 083..1 125] | 1 084 [1 083..1 208] |
| | p99 | 69 183 [67 839..75 967] | 69 375 [68 159..75 263] |
| | p99.9 | 109 375 [106 303..119 359] | 110 591 [108 415..164 223] |

**`reserve_sweep_hdr`** (five scenarios, 9 paired runs each):

| scenario | quantile | `main` | branch |
| --- | --- | --- | --- |
| `nonauto` (every maker strandable; capture + exclusive gate) | p50 | 83 | 83 |
| | p99 | 4 335 [4 001..6 167] | 7 043 [5 711..8 711] |
| | p99.9 | 5 711 [5 003..29 759] | 10 295 [8 295..20 127] |
| `auto` (hidden depth, nothing strandable; count stays 0) | p50 | 833 [666..1 166] | 958 [708..959] |
| | p99 | 3 793 [3 541..12 375] | 3 917 [3 541..4 583] |
| | p99.9 | 5 503 [4 543..32 175] | 5 419 [4 711..8 543] |
| `dense_nonauto` (64 strandable makers on one level, 16..=96 per probe) | p50 | 21 375 [8 295..24 879] | 15 127 [12 047..27 967] |
| | p99 | 56 543 [22 127..158 079] | 37 151 [30 047..70 783] |
| | p99.9 | 68 223 [40 127..1 936 383] | 64 319 [41 631..80 127] |
| `mixed_armed` (one strandable maker parked far away; iceberg levels swept) | p50 | 1 167 [958..1 375] | 1 167 [1 000..1 334] |
| | p99 | 6 003 [4 875..6 627] | 5 751 [5 083..6 751] |
| | p99.9 | 8 543 [7 335..14 671] | 7 795 [6 627..20 015] |
| `mixed_disarmed` (that maker cancelled before the first probe) | p50 | 1 166 [959..1 417] | 1 167 [958..1 416] |
| | p99 | 5 711 [4 959..7 003] | 5 835 [4 751..6 711] |
| | p99.9 | 8 215 [6 751..11 463] | 8 127 [6 127..11 007] |

Reading: a book holding strandable makers pays the capture pass and the exclusive gate, about +2.7 µs at p99 and +4.6 µs at p99.9 on the thin strandable workload; `auto`, `mixed_armed` and `mixed_disarmed` are unchanged within their run ranges, the disarmed scenario confirming that the exact count closes the gate once the last strandable maker is cancelled; `dense_nonauto` measured faster on the branch at p50 and p99, which the added work cannot explain, so no improvement is claimed, only that a dense strandable level is not slower; the common path and `add_only` are unchanged. The capture pass exists because pricelevel drops the stranded hidden quantity without reporting it; having pricelevel report it (or expose a priority-ordered, bounded walk) would remove the pass and is the proposed follow-up in the PriceLevel repo.

## Public API impact

- Removed: `OrderBook::get_bids`, `OrderBook::get_asks` (#228, breaking; migration table in `CHANGELOG.md`).
- Narrowed: the `new` constructors of the `orderbook::iterators` types are `pub(crate)` (#228, breaking in signature only; their argument type was unobtainable through the public API).
- Added: `OrderBookError::ZeroVisibleTranche { order_id, hidden_quantity }` (maps to `RejectReason::InvalidQuantity`), `OrderBookError::ReserveResidualWouldBeDiscarded { order_id, visible_quantity, crossable_quantity, hidden_quantity }` and `RejectReason::ReserveResidualWouldBeDiscarded = 14` (#230, additive; `RejectReason` is `#[non_exhaustive]`).
- Added under the `metrics` feature: counters `orderbook_reserve_discards_total` and `orderbook_reserve_hidden_discarded_total`, recorder `metrics::record_reserve_hidden_discarded` (no-op without the feature).
- `OrderBookError::ReserveResidualWouldBeDiscarded` carries `discarded_quantity` in addition to `hidden_quantity` (review round, unreleased variant).
- Behaviour: in a book holding a non-auto reserve with hidden depth, matching-capable submits and cancel-then-add re-prices take the exclusive submit gate (serialized, as STP books since #225); admitting such a reserve is exclusive in every `STPMode`. `ZeroVisibleTranche` applies to that reserve shape only.
- Version 0.13.0; `src/lib.rs` gains the 0.13.0 What's New section and `README.md` is regenerated via `make readme`; `CHANGELOG.md` carries the breaking subsection and the `### Added` / `### Fixed` bullets.

## Testing

- [x] Unit tests added or updated
- [ ] Round-trip serde tests for new event / DTO shapes
- [x] Integration tests under `tests/unit/` added or updated (if applicable)
- [ ] Snapshot round-trip covered (for changes to `OrderBook<T>` config fields)
- [ ] Replay determinism covered via `snapshots_match` (for journal / sequencer changes)
- [x] Benchmark added/updated (for perf changes): `reserve_sweep_hdr` (HDR harness, two scenarios), wired into `bench-hdr` and `BENCH.md`
- [x] `make pre-push` run locally and clean

#230, new `tests/unit/reserve_residual_policy_tests.rs` (23 tests, public API only) plus `tests/unit/reserve_residual_replay_tests.rs` (1 test), an aggressive 10 / 20 reserve buy against resting sell depth (`replenish_threshold` 0 unless stated). In every case `executed` is the sum of the returned trades and is asserted equal to the contra depth consumed (`seeded - remaining`) and to the tracker's `filled_quantity`; `discarded` is a per-case literal, never derived by subtraction, so `30 == executed + resting visible + resting hidden + discarded` can fail; the engine-side discarded quantity is asserted through the two counters in `tests/metrics/metrics_tests.rs`, the only binary that installs a `metrics` recorder:

| amount / auto | contra | executed | rests | discarded | status |
| --- | --- | --- | --- | --- | --- |
| `None` / off | 5 | 5 | 5 / 20 | 0 | `PartiallyFilled` on the submitted total (30, 5) |
| `None` / off | 10 | 10 | none, order ends | 20 | `Filled { 10 }` |
| `Some(10)` / off | 10 | 10 | none, order ends (the old 10 / 10 residual is gone) | 20 | `Filled { 10 }` |
| `None` / off | 15 | 15 (the taker sweeps with its total of 30) | none, order ends | 15 | `Filled { 15 }` |
| `None` / off | 30 | 30 | none | 0 | `Filled { 30 }` |
| `None` / on | 10 | 10 | 20 / 0 (`min(80, 20)`) | 0 | `PartiallyFilled` (30, 10) |
| `None` / on | 5 | 5 | 5 / 20 | 0 | `PartiallyFilled` (30, 5) |
| `Some(10)` / on | 10 | 10 | 10 / 10 | 0 | resting |
| `Some(10)` / on | 15 | 15 | 10 / 5 (`min(10, 15)`) | 0 | resting |
| `Some(10)` / on, threshold 5 | 8 | 8 | 12 / 10 (visible 2 < 5 triggers `min(10, 20)`, pricelevel's below-threshold rule) | 0 | resting |
| `Some(10)` / off, threshold 5 | 8 | 8 | 2 / 20 (no refresh with automatic replenishment off; visible > 0, nothing discarded) | 0 | resting |
| `Some(10)` / on, threshold 5 | 3 | 3 | 7 / 20 (visible 7 >= 5, the arm does not fire) | 0 | resting |

Zero-visible rule (4 tests): a reserve and an iceberg submitted as `{0, 20}` are rejected with `ZeroVisibleTranche { hidden_quantity: 20 }` and create no level; a standard order with quantity 0 is not touched by the rule; on a resting reserve and a resting iceberg, `UpdateQuantity`, `UpdatePriceAndQuantity` and `Replace` with quantity 0 are rejected and leave price, tranches and level untouched. Modify pre-check (6 tests, each looping `UpdatePrice`, `UpdatePriceAndQuantity` and `Replace`): a resting non-auto 10 / 20 reserve re-priced into contra depth 10 or 15 is rejected with `ReserveResidualWouldBeDiscarded` and left untouched, into depth 5 executes 5 and rests, into depth 30 or more fills completely, and the auto-replenishing twin is never rejected. `reserve_residual_replay_tests.rs` journals the aggressive `{10 / 20, None, off}` session against contra depth 10 as `SequencerCommand::AddOrder` events in an `InMemoryJournal`, replays it with `ReplayEngine::replay_from_with_clock` under a `StubClock`, and asserts the replayed book ends the order (no resting residual, both sides empty, last trade at the recorded price) and is `snapshots_match`-equal to the live book that produced the same two submits; nothing under `src/orderbook/sequencer/` changes.
`tests/metrics/metrics_tests.rs` gains the engine-side measurement on both paths: the taker discard moves `orderbook_reserve_discards_total` by 1 and `orderbook_reserve_hidden_discarded_total` by 20, the maker removal (resting non-auto 10 / 20 hit by a sell of 20) moves them by the same amounts, and the auto-replenishing twin moves neither.

Plus the execution-asymmetry test: the same `{10 / 20, None, off}` order resting as a maker hit by an aggressive sell of 20 executes only 10, leaves the book as `Filled { 10 }` with 20 discarded, while the aggressive version above executes 20 against 20 of liquidity; and, after every non-auto termination, no order and no level remain (`get_order` `None`, `best_bid` `None`).

Migrated rather than deleted: `test_validate_order_shape_iceberg_and_reserve_share_tranche_verdicts` sweeps the `(0, 20)` pair, which now returns `ZeroVisibleTranche` for both kinds; the test asserts both name the same hidden tranche, keeping the symmetry claim. The #226 lot-size tests follow the gated transfer rule: `test_validate_order_shape_reserve_misaligned_replenish_amount_without_auto_accepts` (10 / 20, `Some(7)`, off) now accepts while the `auto = true` sibling rejects; `test_add_order_reserve_misaligned_replenish_amount_with_auto_rejects` (renamed, `auto = true`) gained a `..._without_auto_accepts` twin; `test_reserve_taker_misaligned_replenish_amount_rejects_before_matching` uses `Some(7), true`; the aligned-residual case uses `Some(10), true` (rests 10 / 10 on a lot-10 book), the non-auto termination having moved to the new file.

#228 has no runtime test to add: the accessors had no callers; the removal is verified by the build and by the `semver` job accepting the 0.13.0 bump.

Co-located `src/orderbook/tests/strandable_maker_count.rs` (15 tests) pins the capture gate and the exact count: false at construction, stays false for iceberg-only and auto-reserve books through a sweep and for a non-auto reserve without hidden depth, set by a resting non-auto reserve with hidden depth and by a rested residual, re-derived by `restore_from_snapshot_package` (and not set by an iceberg-only package), and a closed flag never changes a matching outcome.

`src/orderbook/tests/strandable_gate_concurrency.rs` (3 tests) holds the deterministic interleavings described in the review-round section; `src/orderbook/tests/error.rs` and `reject_reason.rs` gain `Display` / `Clone` / mapping / wire-code tests for the two new variants; the STP `CancelTaker` capture site has a direct test.

Suite: 863 lib (1 ignored) + 1 alloc_budget + 26 fee + 6 metrics + 560 integration + 49 doctests, 0 failed (`cargo test --all-features`); `make test` (default features) 730 + 26 + 522 + 46, 0 failed. `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo build --release`, `make doc` clean; `make readme` regenerates `README.md` with only the new 0.13.0 section (+107 lines) and is idempotent.

## Checklist

- [x] Module boundaries respected (core engine has no deps on `manager`, `nats*`, `sequencer/`)
- [x] No new dependencies added without explicit approval
- [x] `src/lib.rs` docs + `README.md` (via `make readme`) + `CHANGELOG.md` updated if user-visible

Closes #228
Closes #230

https://claude.ai/code/session_017FAapGt6dfZYjHMPUngzYZ


src/orderbook/matching.rs
```diff
@@ -402,8 +402,28 @@ where
...
+        // Same `(Id, quantity)` shape as `filled_orders`, so it reuses the
+        // same pool.
+        //
+        // The gate is read ONCE here, not per level: on a book that has never
+        // rested a non-auto-replenishing reserve with hidden depth the whole
+        // strandable path drops out of this sweep — `None` means no pool
+        // acquire, no per-level capture, no lookup in the drain loop and
+        // nothing to return. That matters because the capture walks the
+        // level's `DashMap` of orders, which read-locks every shard.
+        let watch_strandable = self.non_auto_reserve_rested.load(Ordering::Relaxed);
```
joaquinbejar: 🟠 P2 — Reading this flag once can silently lose a discard report. Deterministic reproduction under STPMode::None: rest ordinary asks 1@100 and 1@101; start a buy of 3@101; pause after the first level fills; on another thread admit the first non-auto reserve (visible 1, hidden 20) at 101; resume. The reserve is consumed and removed, but the discard-quantity counter remains 0 instead of 20, and the maker INFO report is skipped because strandable_makers stays None for the entire sweep. The same interleaving passes if a previously cancelled reserve prearmed the flag. This uses public APIs; a listener/barrier only fixes the schedule. Make admission, capture, and matching preserve a coherent reporting boundary. Switching Relaxed to Acquire or moving the store alone cannot fix a value that becomes stale after the load; capture and match are also separate operations under the shared gate.

src/orderbook/matching.rs
```diff
@@ -699,13 +728,35 @@ where
...
             self.track_state(
                 *filled_id,
                 OrderStatus::Filled {
                     filled_quantity: *filled_quantity,
                 },
             );
+            if let Some(strandable) = strandable_makers.as_ref()
+                && let Some((_, discarded_hidden)) = strandable
+                    .iter()
+                    .find(|(strandable_id, _)| strandable_id == filled_id)
```
joaquinbejar: 🟠 P2 — The new reporting drain adds O(F × S) work to matching: each of F filled makers searches all S accumulated captured candidates. Consuming N strandable makers therefore adds quadratic identity comparisons, on top of the full-level DashMap capture pass. This cost is not bounded by the current three-maker reserve_sweep benchmark. Also, cancelling the last non-auto reserve never clears the flag, so later iceberg/auto-reserve levels still pay the scan. Avoid the repeated linear search, and add dense-level plus previously-flagged mixed-book measurements before treating the reported ~2.4 µs thin-book p99 delta as the accepted cost. This concerns the new reporting work; it is not a claim that all pre-existing matching work was linear.

src/orderbook/modifications.rs
```diff
@@ -1359,6 +1524,99 @@ where
...
+        // between destroys quantity. `crossable == 0` is defense in depth:
+        // `validate_order_shape` already rejects a projected zero visible
+        // tranche, and without that rule a non-crossing re-price of such an
+        // order would fall inside the band vacuously.
+        if crossable > 0 && crossable >= visible && crossable < total {
+            return Err(reserve_residual_would_be_discarded(
+                new_order.id(),
+                visible,
+                crossable,
+                hidden,
```
joaquinbejar, line 1667: 🟡 P3 — The error reports the submitted hidden tranche as the amount that would be discarded. For visible 10 / hidden 20 / crossable 15, this passes 20, and Display says 20 hidden units would be discarded. An explicit cancel-and-submit probe executes 15 and discards only 15 (total 30 minus executed 15). The rejection band is correct; its diagnostic is not. Either carry the predicted remainder `total - crossable`, or clearly distinguish original hidden quantity from predicted discarded quantity in the field docs and Display. Add a diagnostic assertion where crossable is strictly greater than visible.

src/orderbook/modifications.rs
```diff
@@ -1123,6 +1262,30 @@ where
...
+            ..
+        }
+        | OrderType::ReserveOrder {
+            visible_quantity,
+            hidden_quantity,
+            ..
+        } = order
+            && visible_quantity.as_u64() == 0
+            && hidden_quantity.as_u64() > 0
+        {
```
joaquinbejar: 🟠 P2 — Separate this broader admission policy from the claimed phantom-order bug. The comment and public ZeroVisibleTranche docs say every zero-visible two-tranche order is unfillable and loses its hidden depth; that is false for icebergs and auto-replenishing reserves. On this head, a restored legacy auto reserve 0/20 successfully executes 10, and this new gate also rejects an aggressive auto reserve 0/20 crossing 20 available units even though the established taker path matches total quantity. The non-auto residual fix does not itself require rejecting those executable orders. Either narrow the rule or explicitly retain it as a new positive-display admission restriction with an accurate compatibility rationale. Keep the #223 zero-UpdateQuantity cancellation exception distinct during integration.

doc/USER_GUIDE.md
```diff
@@ -151,13 +151,42 @@ println!("Average price: {}", result.average_price());
...
+  no depth, never fill, and lose its whole hidden tranche to the first taker
+  that reached its level
+- An aggressive two-tranche order sweeps with its **total**, not with its
+  visible tranche: a 10 visible / 20 hidden Reserve submitted into 20 units
+  of contra liquidity executes 20. What its unmatched residual does follows
+  `auto_replenish`. With it on and hidden left, a visible tranche the sweep
+  left below `max(replenish_threshold, 1)` — an emptied one always is — is
+  refreshed from hidden (the explicit `replenish_amount`, or
+  `DEFAULT_RESERVE_REPLENISH_AMOUNT` when there is none, capped by hidden)
+  and the residual rests. With it off the residual does not rest at all: its
```
joaquinbejar: 🟠 P2 — Two policy statements here contradict the implementation and the PR's own tests. (1) With auto_replenish=false, the residual is discarded only when the visible tranche is exhausted: 10/20 filled for 5 rests as 5/20. Add that condition to this paragraph and the parallel add_order/lib.rs/README wording. (2) An explicit replenish_amount controls the transfer, not a cap on displayed size: amount 10, threshold 5, and visible remainder 2 produce visible 12 after replenishment. Replace the 'pin the displayed size' advice at line 189 and its parallel API/release wording with that precise behavior.

joaquinbejar (commented):
Reviewed exact head 25474ca independently, including core, matching/observability, persistence, API boundaries, and architecture. The focused reserve suite (50 tests), rejection taxonomy (23 tests), three shipped discard-metric tests, replay/restore probes, and a 7,360-case residual-helper versus upstream comparison pass. The #228 cleanup is sound: remaining get_bt_bids/get_bt_asks reconstruct detached levels, iterator internals stay private, and 0.13.0 covers the removals. Snapshot flag reconstruction and wire code 14 also check out.

I would hold the merge for the new discard-reporting race and the reporting loop's scaling issue below. A deterministic public-API interleaving drops 20 hidden units while reporting zero; prearming the flag makes the same test pass. The thin three-maker benchmark does not cover the added O(filled makers × captured candidates) lookup cost or a previously flagged book now holding only other hidden-order types.

The remaining points concern the inaccurate discard diagnostic, the broader positive-display admission policy, and public documentation. In addition to the inline doc corrections, qualify src/lib.rs's absolute 'a modify can never silently destroy the order' statement with the shared-gate/best-effort limitation already documented on check_modify_reserve_residual, then regenerate README. These corrections should precede ratification of the proposed observability and performance tradeoff.

src/orderbook/matching.rs
joaquinbejar: Right, the per sweep read could go stale under the shared gate. Every sweep entry point in a book that holds strandable makers now runs on the exclusive side (decided from the resting count, re-checked after acquiring the shared side and re-acquired exclusive from scratch if it changed, nothing read before), and admitting a strandable maker is exclusive as well, so no cancel, mass cancel, admission or re-price can land inside a level's capture window. Your interleaving and the id reuse one (through add_order and through the anonymous match_order) are pinned in strandable_gate_concurrency.rs, pre-fix failures in the PR body.

src/orderbook/matching.rs
joaquinbejar: The drain now sorts the captured makers once and binary searches per filled id, and the flag became an exact count of resting strandable makers, decremented from the removed order's own body on cancel, every mass cancel entry point, expiry, STP removal and fill, reset on cancel_all_orders and restore, with a test per path plus id reuse and cancel versus match. reserve_sweep_hdr gained dense, mixed armed and mixed disarmed scenarios, paired main versus branch numbers in BENCH.md and the PR body.

src/orderbook/modifications.rs
joaquinbejar: Good catch. The error now carries discarded_quantity, the remainder total minus crossable that the re-add would actually destroy, next to the projected hidden tranche, Display distinguishes the two, and the modify test asserts 15 discarded of 20 hidden at depth 15.

src/orderbook/modifications.rs
joaquinbejar: Narrowed to the non auto reserve with visible zero and hidden positive, the only shape pricelevel removes without a trade. Zero visible icebergs and auto reserves are admitted again and covered as makers and as partially filling takers, and the #223 zero UpdateQuantity exception is documented as taking precedence for that variant.

doc/USER_GUIDE.md
joaquinbejar: Both fixed in the guide, set_total_remaining, add_order, lib.rs and README. The discard needs the visible tranche exhausted (10/20 filled for 5 rests 5/20), and the explicit amount is the transfer added to whatever visible survives (amount 10, threshold 5, remainder 2 gives 12), the pin the displayed size wording is gone.

src/orderbook/modifications.rs
```diff
@@ -423,7 +521,13 @@ where
                     // BEFORE cancelling the original, so the original survives.
                     self.check_modify_stp_self_cross(&new_order)?;
 
-                    // Both checks passed: cancel the original and add the
+                    // #230: reject a re-price whose re-add would exhaust a
+                    // non-auto-replenishing reserve's visible tranche and
+                    // discard its hidden remainder, which would destroy the
+                    // order after the original was already cancelled.
+                    self.check_modify_reserve_residual(&new_order)?;
```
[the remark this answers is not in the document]
joaquinbejar: The re-price of a strandable reserve now holds the exclusive gate from lookup through validation, cancel and re-add, so the dry run cannot be invalidated by a concurrent opposite side admission. The documented guarantee is scoped to exactly that case.

src/orderbook/matching.rs
```diff
@@ -867,6 +921,60 @@ where
...
+        if !self.non_auto_reserve_rested.load(Ordering::Relaxed)
+            || price_level.hidden_quantity() == 0
+        {
+            return;
+        }
+        for order in price_level.iter_orders() {
+            if let OrderType::ReserveOrder {
+                hidden_quantity,
+                auto_replenish: false,
+                ..
```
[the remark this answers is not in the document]
joaquinbejar: Capture and match can no longer be separated by an admission, a cancel or an id reuse. Every sweep entry point in a book holding strandable makers runs under the exclusive gate and admitting one is exclusive too, with deterministic regressions for both interleavings.

joaquinbejar:
Thanks for the deep pass, all six conditions are in. Sweeps in books holding strandable makers take the exclusive gate with the count based restart pattern, the id reuse interleaving has its regression, the counter is exact with a test per removal path, ZeroVisibleTranche is narrowed with maker and taker coverage, the ghost shape is rejected in prepare_snapshot_levels on all three restore paths, the diagnostic carries the predicted discard and the modify guarantee is scoped. Copilot's two suppressed points (legacy ghost on restore, the band wording) are covered by the same changes. Paired main versus branch numbers for the five reserve_sweep scenarios are in BENCH.md and the PR body.

